PR #2, opened 2021-04-04
Add bash scripts; enhance ASMx, CBLCx bat files; add CBLCx to commands; move COBOL opt to separate directory

jyganci:
Add bash directory that contains bash scripts used by Linux and Mac users to perform same functions as corresponding BAT files.


z390org:
Looks good. You mention a few renames are needed. Are you planning to make that your next submission?

jyganci:
Abe,

I told Don about the 2 renames a few weeks ago. He has already made the
changes to his versions. They will be included in 17_07.

The mention of the renames in the bash readme is there just in case someone
tries to use the bash scripts before v17_07 is released.

z390org:
John,

I understand - the trick is, you *might* have packaged that all
in one submission.

The trick with git/github is that it relieves of the hassle
of coordinating such changes with other contributors.

Especially when you're working with short-lived branches
the risk of conflict is pretty small...

The old-school method still works just as well.
I'm glad it's already been taken care of by Don.

Kind regards,
Abe
===



Op 04/04/2021 om 20:16 schreef John Ganci:

adelosa:
Would we consider moving all the scripts into a **bin** folder and have the windows ones *.bat and the unix ones just with no extension. That way, on either OS, you could add the bin folder to your path and issue the commands without the extension.
I think we should keep them separate rather than having a poly type file as code editors and people may get confused..
The other thing is, would this deprecate the perl scripts? and should they be removed as part of this change..

jyganci:
Since most bash scripts that reside in system bin directories do not have the ".sh" extension, I struggled with whether to remove the ".sh" extension from the script names but decided that, for now, to leave them. I also decided to use "bash" for the directory name, not "bin", at least for the initial installation of the bash scripts. However, I will remove the ".sh" extensions and rename bash to bin if that's what everyone wants. Even though I do not plan to add any z390 directories to my path, I do agree with Anthony's statement about issuing the commands without the extension.

I do not think the bash and bat directories should be combined. Keeping them separate leaves no room for confusion.

We must keep the Perl scripts for now because anyone using the z390 GUI will be using them until the z390.java code is modified. I do plan to look into this, but anyone else is welcome to do so.

Lastly, no one has yet brought up anything about CBLOPT.OPT being in the bash and bat directories. It really belongs somewhere else, but because it's been in the bat directory for a long time, I decided not to tackle that issue at this time. It is a "generic" COBOL options file, not specific to any COBOL program. Any suggestions about what to do about it? Whatever is done, changes will then be required for the COBOL compile(link, link and go) bash and bat files.

bash/asm
```diff
@@ -0,0 +1,46 @@
...
+
+# get the z390 directory
+zdir=$(dirname $0)
+zdir=$(cd $zdir && pwd)
+zdir=$(dirname $zdir)
+
+sysmac='sysmac(+'$zdir'/mac)'
+syscpy='syscpy(+'$zdir'/mac)'
+
+${dir}/mz390 $1 $sysmac $syscpy $2 $3 $4 $5 $6 $7 $8 $9
```
adelosa, line 39: Is it worth moving to the bash way of passing the remaining items to the next command.
* `"$@"` provides all the parameters passed on the command line.. 
* `shift` command remove the variables in the list.

Means you can have more than 2-9 options passed.
Would look something like:

    module=$1
    shift
    mz390 $module $sysmac $syscpy "$@"

jyganci: I'd forgotten about $@. Why the double quotes enclosing $@ on the "mz390 $module ..." line?

adelosa: I am not 100% why.. but all the guides I read said you should.. I think it might be to ensure that bash expands it.

jyganci: While I think using $@ is very cool, I would prefer to leave the scripts as-is. Since the bash scripts are new, any one who wants to compare a bat file to its corrsponding bash script will see similar code.

bat/ASML.BAT
```diff
@@ -1,26 +1,31 @@
...
+if exist %1.OBJ erase %1.OBJ
+if exist %1.LST erase %1.LST
+if exist %1.390 erase %1.390
+if exist %1.ERR erase %1.ERR
+if exist %1.STA erase %1.STA
+if exist %1.TR* erase %1.TR*
+
+rem get the z390 directory
+set "zdir=%~dps0..\"
+for %%f in ("%zdir%") do set "zdir=%%~ff"
```
adelosa: Just so I get this, this command is setting the zdir folder to the parent folder?? So this assumes a specific folder structure when running your assembles..

That is:

    project
    -- code (where the mlc lives)
    -- mac (for any referenced macros or includes)

Just want to understand as I think this should be documented somewhere.. I'm looking at the doco. I assume the

adelosa (commented):
I am personally happy to merge this. I have included a couple of comments for consideration but no changes required. I don't think we have agreed who needs to sign off pull requests before they get merged. I am thinking @z390org/@abekornelis  or @dsh33782 will be providing the actual approval. 
Once that is done, we can merge unless there is something else you intend on adding to the pull request.

bat/ASML.BAT
jyganci: This may be more than you wanted to know.

Yes, the z390 software assumes a particular directory hierarchy (the hierarchy used in github). The z390 bat files are assumed to be in the bat subdirectory of the z390 root directory, z390 macros in the mac subdirectory, etc. The variable %zdir% is first set to a relative path for one directory up; the "for" statement converts it to a full path name. A user's mlc files would NOT reside within the z390 root directory. For example, if user fred has file C:\fred\source\MYPGM.MLC, fred could assemble and link it using
    cd <full path to the z390 root directory>
    bat\asml C:\fred\source\mypgm

Line 20 of ASML.BAT

call %~dps0mz390 %1  sysmac(+%zdir%mac) syscpy(+%zdir%mac) %2 %3 %4 %5 %6 %7 %8 %9

will correctly set the sysmac and syscpy values to full path names to the z390 mac directory prefixed with "+", which is a signal to concatenate. (If I remember correctly, the full path directory name for %1 will be the first part of the final sysmac value used, followed by the above value. Abe can keep me honest here.)

You will also notice that line 20 prefixes mz390 with %~dps0, so z390 assumes that MZ390.BAT is in the bat directory.

Looking in MZ390.BAT, line 8, (shortened here)

java -classpath %~dps0..\z390.jar -Xrs -Xms150000K  ...

note the classpath is set to the z390.jar file, which is assumed to be in the z390 root directory. Another use of moving one directory up.

These assumptions (z390.jar in z390 root directory, names and locations of subdirectories) are used throughout the bat files and the newer bash files.

While a developer might be concerned about these assumptions, I'm not sure a user would be.

Hope this answers your questions.

On another note, I'd like to learn how you included the commented ASML.BAT lines.

jyganci:
I hope that this last commit, add cblc, cblcl, cblclg to commands, is the last commit for this pull request.

abekornelis (approved):
I find this hard to assess. Looks good to me though, for what it's worth.

abekornelis:
Approved by me and don Higgins

PR #264, opened 2021-05-21
V1707fixes

dsh33782:
This pull request is to merge 3 changes to the V1707 branch including update of bat\Z390.BAT, docuement addition of 2 more directories from v1706 that were not in the v1707zip, and not V1707 repository does not have z390.jar which will require user to install java 8.0+ jdka and then run bat\BLDJAR.BAT to create it.


abekornelis (approved):
Looks good to me

PR #269, opened 2021-05-25
Add missing demo and zpar files to v1707

dsh33782:
This pull request adds missing files for demo and zpar directories omitted from v1707 that were in v1706 repository.


abekornelis (commented):
I see there's a tempgen.bat.
Are we sure this file belongs in the repo? 
If yes, shouldn't it have a more descriptive name?

dsh33782:
Abe

 

I’ll delete it as it is a zpar test of all the NIST COBOL 85 test programs that does not

Belong in zpar dir.

 

Don Higgins

 ***@***.***> ***@***.***

 <http://www.don-higgins.net/> www.donhiggins.org

jyganci (approved):
Only did a scan of the files. Approved.

PR #297, opened 2021-06-02
V1600next

### LINKED ISSUE #255: Add missing improvements from the V1.6.00 beta series back in
Code changes realized after publishing beta V1.6.00b12 have never made it into the current V1.7 series of distributions. We need to add these la

abekornelis:
These are the left-over changes from V1.6.00b13 which has never been distributed.
Please review changes. 
There's one thing that worries me: it appears the zcobol/mac/gen_div.mac lost changes due to RPI 1217, which was coded by Don.

this pull request should fix #255 Add missing improvements from the V1.6.00 beta series back in

The majority of changes in this request were authored by John Hennesy and relate to zcobol support.
There is one apparent regression issue: in zcobol/mac/GEN_DIV.MAC it looks like RPI 1217 was undone.
Looking at the history, it seems the following happened:
- Don fixed RPI 1211 on May, 2nd, 2012 = add support for implied decimal points
- Don fixed RPI 1217 on May, 18th, 2012 = replace SCALE_F with SCALE_G to prevent F/H ovf.

- John fixed RPI 1559 on January, 22nd, 2017 = Error in GEN_DIV where &LNQ should be &LQ
- John fixed RPI 1560 on January, 22nd, 2017 = Error in GEN_DIV where &LNR should be &LR (two locations)

The version with the two fixes by Don appears in distributions V1.5.06 through V1.6.00b12

The version with the fixes by John appears without the RPI 2017 fix.
Most likely John took an intermediate version from V1.5.06rc4 which was the last rc version before Don released V1.5.06.
This version did have the RPI 1211 fix but not the RPI 1217 fix.

@dsh33782 and/or @John_Hennesy would you two please have a look?
I'd like to have a consolidated version that has all four fixes in it before proceeding with this pull request.


abekornelis:
Don commented by email:
Upon further research, The above structured macro assembler program DEMOM81.ZSM will assemble without errors using the mz390.java macro processor via bat\ASM.BAT DEMOM8Q1.ZSM command. All the structured macro support implemented in the ZSTRMAC.MLC translator has been added to the mz390.java macro processor. The only reason to use the ZSTRMAC.MLC translator would be to convert structured macro source programs to BAL for assembly using HLASM assembler.

I propose to replace the original demo\DEMOM8Q1.MLC with the newer more structured version above which has no macro labels or branch statements in it. Both versions produce the same solution as PRN output.

I recommend using the same newer source program as rt test input to the ZSTRMAC.MLC translator to verify the translated version produces the same solution.

The documentation on z390 structured macro extensions including ZSTRMAC.MLC translator are currently located here:

http://www.z390.org/z390_ZSTRMAC_Structured_Macro_Support.htm

abekornelis:
Don commented by email:
Regarding the RPI 1217 changes to GEN_DIV.  I did a search of z390 repository for RPI 1217 and found changes for the following zcobol macros:  ZC_UNPK31, GEN_COMP, GEN_MPY, GEN_ADD, GEN_SUB, GEN_DIV.  The comment description for RPI 1217 that perhaps best fits is the following:

   05/18/12 RPI 1217 replace SCALE_F with SCALE_G to prevent F/H of.

My recollection is a little fuzzy after 9 years, but I think the issue was overflow when scaling values using 32 bit versus 64 bit work registers.

abekornelis:
Relating to the issue on the GEN_DIV macro I replied:
Don, John, all,

I've made an attempt at creating a consolidated version of GEN_DIV.MAC
containing both the two fixes by Don (RPI 1211 + 1217)
and the two fixed by John (RPI 1559 + 1560).

I've attached the result to this email.

If/when you can find the time would you please have a look
to double-check whether this seems okay?

I understand it may take a couple of days before either
of you finds an opportunity to spend some time on this.

The pull request with this (and other) changes will have
to remain on hold until this has been sorted out.

Man - am I glad we're using git!
Without the git tooling I could never have come this far.

[GEN_DIV.TXT](https://github.com/z390development/z390/files/6592688/GEN_DIV.TXT)

Then Don replied:
The changes to GEN_DIV look ok to me.  

I think the entry for RPI 1217 could be removed as there is no scale adjustment or operands required for divide like there is for add/sub.  The final implied decimal point is only adjustment required.

To which I in turn reply:
RPI 1217 has seen extensive changes to use 64-bit registers instead of 32-bit registers. It seems the intention was to avoid overflow problems. I would be hesitant to discard these changes without proper testing. RPI 1217 was created for a reason, even if nobody recalls all the details of the how and why. I'm inclined to carry on with the integrated version. If any problems arise we can raise a new issue.

jyganci:
Finished looking through the changes and the GEN_DIV.MAC sent in separate email. All looks good to me. Even though it does not affect the pull request, I would like to hear back from Don regarding the questions I posed about ZCVTD.MAC.

So, is it time to do the review and approval so the merge can be done?

abekornelis:
John, 

I'm inclined to move ahead - but let's wait and give Don the opportunity to have the last word on GEN-DIV.MAC.
unless he objects, I'd say you should be fine to execute the merge by the end of today - U.S. time ;-)

jyganci (approved):
Looks good. Approved.

Issue #291, opened 2021-05-31
ASSIST XDECI behaviour not correct


adelosa:
In reviewing the documentation for z390 ASSIST, the function was compared to the original ASSIST assembler doco.

> XDECI is generally used to scan a data card read by XREAD. The sequence of actions performed by XDECI is as follows:
> 1. Beginning at the location given by ADDRESS, memory is scanned for the first character which is not a blank.
> 2. If the first character found is anything but a decimal digit or plus or minus sign, register 1 is set to the address of that character, and the condition code is set to 3 (overflow) to show that no decimal number could be converted. The contents of REG are not changed, and nothing more is done.
> 3. From one to nine decimal digits are scanned, and the number converted to binary and placed in REG, with the appropriate sign. The condition code is set to 0 (0), 1 (-), or 2 (+), depending on the value just placed in REG.
> 4. Register 1 is set to the address of the first non-digit after the string of decimal digits. Thus REG should not usually be 1. This permits the user to scan across a card image for any number of decimal values. The values should be separated by blanks, since otherwise the scanner could hang up on a string like -123*, unless the user checks for this himself. I.e. XDECI will skip leading blanks but will not itself skip over any other characters.
> 5. During step 3, if ten or more decimal digits are found, register 1 is set to the address of the first character found which is not a decimal digit, the condition code is set to 3, and REG is left unchanged. A plus or minus sign alone causes a similar action, with R1 set to the address of the character following the sign character.

* Item 3 is not aligned. The z390 implementation does not set the condition code based on the sign of the result. If converted, it always sets the condition code to zero.
* Item 5 is not aligned. The z390 implementation will continue to convert after 9 digits and does not set CC=3.

I wrote the following program to test the behaviour as described above. Obviously the ASSIST option needs to be enabled to run this program.

My view is the following program should run successfully if the XDECI instruction has been implemented correctly.
**NOTE**: I am a beginner assembler programmer and this is probably not how an experienced programmer would create the program but its what I know. Any ideas on improvements would also be appreciated. 

The fix should include the test program as part of the regression test suite.
 
```
* 
* ASSIST1 - Tests ASSIST XDECI function
*
ASSIST1  SUBENTRY
         USING TESTDATA,6      R6 to use testdata dsect
         LA    7,NUMTESTS      R7 contains test counter
         LA    6,TEST1         R6 points at current testdata
LOOP     PERFORM CHKXDECI      Perform the test
         LA    6,TDATALEN(,6)  Increment R6 to next test
[...]

PR #308, opened 2021-06-09
Issue#291: Modify XDECI code to match ASSIST doc

jyganci:
While creating the markdown documentation, it was discovered that the z390 implementation of the ASSIST XDECI instruction had several problems. These problems  are corrected in this pull request.

In addition, a new test program, TSTXDECI.MLC, which is based on the ASSIST1.MLC test program that is shown in issue #291, is being added to the assist/test directory. It tests these changes. Run with
    Windows: bat/ASSIST.BAT assist/test/TSTXDECI
    Linux/Mac: bash/assist assist/test/TESTXDECI

### DIFF
--- a/assist/test/TSTXDECI.MLC
+++ b/assist/test/TSTXDECI.MLC
@@ -1,0 +1,345 @@
+**********************************************************************
+*
+* TSTXDECI: Tests ASSIST XDECI instruction
+*
+* All tests should succeed.
+*
+**********************************************************************
+*
+TSTXDECI CSECT
+         STM   14,12,12(13)    Save caller's registers
+         LR    12,15           R12 = base register
+         USING TSTXDECI,12     Establish addressability
+         LA    14,SA           R14 --> usable save area
+         ST    14,8(,13)       Set caller's forward chain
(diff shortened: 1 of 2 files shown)


dsh33782 (approved):
Looks ok to me.  Wish I could test on Windows before approving but I can't right now.

jyganci:
Don,

Attached is a zip file containing
1. The updated pz390.java
2, The test program I used TSTXDECI.MLC
3. Output I see when running the test

The test program does eleven tests. For each test, there are three checks
done after XDECI is issued:
1. Is the condition code set correctly?
2. Is register 1 set correctly?
3. Is the output register set correctly?

If any of these is "No", an error message is written.

If you are so inclined, you can run the tests yourself:
1. Back up your existing pz390.java and z390.jar
2. Copy my pz390.java to your z390 src directory
3. Rebuild the jar
4. Run the test program using bat/ASSIST.

As you can see in the sample output, I ran my test when positioned in the
main z390 directory.
The test program TSTXDECI.MLC is in another directory, so I had to provide
the full path
to it.

When done testing, delete the new pz390.java, restore your backed up
pz390.java and z390.jar.

PR #309, opened 2021-06-09
fix issue 244 by updating bat\RUNVSEDEMOS.BAT

dsh33782:
Fix issue 244 by updating bat\RUNVSEDEMOS.BAT to include call before each bat\ASMLG command.  The previous bat fiel was terminating after first demo without pause to verify and without continuing on to demos 2 and 3.  All work correctly now.  The original issue was raised on v1707 where vse assembly was resulting in base register error.  John fixed that error and successfully tested the vse demos using bash scripts on Linux.


jyganci (approved):
Looks good. Approved.

PR #313, opened 2021-06-12
Automated build process

### LINKED ISSUE #312: Automated build process - create the jar
Please set up github to execute the bldjar procedure

adelosa:
This PR covers the linux build and release artefact.

## Build process

Triggered when code pushed to the repo.
Runs build.sh in root folder. 
See https://github.com/adelosa/z390/actions/runs/1282628473 for results.

Additional actions can be executed by adding them to the build.sh file in the root.

## Bash script changes

Standard bash scripts were changed due to a couple of issues:

1) bldjar would just fail if the jar folder did not exist. The amended script will create if it does not exist.
2) Pauses or user input requirement removed. Just runs. You can add back but you will need to make it an option rather than the default
3) Scripts must be run with shell in the bash folder - changed so can be run anywhere

See commit 675a1e4 for the changes made to the bash scripts.

## Dynamic versioning

I have changed Java program TZ390 to load a properties file to determine the version. The z390.properties file is generated as part of the bldjar script and injected into the jar file. If it can't find the properties file, it will still compile, but the version will show as "Uncontrolled". We use git to provide a version based on the tags applied. No more hard coded versions..

```
13:07:24 IEFBR14   MZ390 START USING z390 v1.7.07-113-g1c317ec ON J2SE 1.8.0_302 09/25/21
13:07:24 IEFBR14   MZ390 ENDED   RC= 0 SEC= 0 MEM(MB)= 42 IO=32
```

The version in this case is v1.7.07-113-g1c317ec - It is based on the last git tag - number of subsequent commits - commit sha1

## Create Distribution file

A new bash/blddist script creates a distribution zip file after the linux build completes. 

The distribution file will have the name z390-v1.7.07-113-g1c317ec.zip - the version is part of the name.

## Create GitHub release

The distribution file is stored in GitHub as a releases artefact. See my clone for what this looks like.

https://github.com/adelosa/z390/releases

The release is automatically created when a git release tag (starts with 'v') is added to the repo.

## Distribution documentation

I will not be initially including documentation into the distribution. I have another PR called "doc structure" which I need to merge before we have a good working state and can update it to add the docs to the distribution. 

## NEXT STEPS

All comments welcome. Try the generated distribution and see if it causes any issues on your system.

* Release page: https://github.com/adelosa/z390/releases/tag/v1.7.08
* Release download: https://github.com/adelosa/z390/releases/download/v1.7.08/z390_v1.7.08.zip


dsh33782 (dismissed):
Great first version that will be very useful for testing changes.  For next version I'd like to see addition of command to verify that selected output file contains specified text etheir at specific offset or anywhere.  For example in RUNASMDEMOS.BAT add command to verify HELLO.LOG file contains "Hellow World".

dsh33782:
Great first step at rt for builds

bash/bldcbllib
```diff
@@ -1,29 +1,18 @@
...
-${dir}/bash/asm zcobol/lib/INSPECT $sysmac $syscpy $1 $2 $3 $4 $5 $6 $7 $8 $9
-${dir}/bash/asm zcobol/lib/ZC390NUC $sysmac $syscpy $1 $2 $3 $4 $5 $6 $7 $8 $9
-${dir}/bash/asml zcobol/lib/ZC390LIB $sysmac $syscpy $sys390 RMODE24 $1 $2 $3 $4 $5 $6 $7 $8 $9
+bash/asm zcobol/lib/ABORT $sysmac $syscpy $1 $2 $3 $4 $5 $6 $7 $8 $9
+bash/asm zcobol/lib/ACCEPT $sysmac $syscpy $1 $2 $3 $4 $5 $6 $7 $8 $9
+bash/asm zcobol/lib/CVTTOHEX $sysmac $syscpy $1 $2 $3 $4 $5 $6 $7 $8 $9
+bash/asm zcobol/lib/DISPLAY $sysmac $syscpy $1 $2 $3 $4 $5 $6 $7 $8 $9
+bash/asm zcobol/lib/INSPECT $sysmac $syscpy $1 $2 $3 $4 $5 $6 $7 $8 $9
+bash/asm zcobol/lib/ZC390NUC $sysmac $syscpy $1 $2 $3 $4 $5 $6 $7 $8 $9
+bash/asml zcobol/lib/ZC390LIB $sysmac $syscpy $sys390 RMODE24 $1 $2 $3 $4 $5 $6 $7 $8 $9
```
jyganci, line 19: Since you deleted the setting of sysmac, syscpy, and sys390, aren't they null in the replaced lines? Is this intentional?

bash/runasmdemos
```diff
@@ -1,30 +1,22 @@
...
+set -e              # exit on all errors
+cd $(dirname $0)    # set to directory of script
 cd ..
 
-${dir}/bash/asmlg demo/HELLO $1 $2 $3 $4 $5 $6 $7 $8 $9
+bash/asmlg demo/HELLO $1 $2 $3 $4 $5 $6 $7 $8 $9
 echo "Verify Hello World WTO"
 
-${dir}/bash/asm demo/DEMOM8Q1 $1 $2 $3 $4 $5 $6 $7 $8 $9
+bash/asm demo/DEMOM8Q1.ZSM $1 $2 $3 $4 $5 $6 $7 $8 $9
```
jyganci: I don't remember DEMOM8Q1.MLC being replaced by DEMOM8Q1.ZSM, but okay, I guess.

bash/bldcbllib
adelosa: That is interesting.. I probably did that accidentally, but it still works.
The asm and asml scripts already do a sysmac and syscpy include, but it does not zcobol/lib include.
Maybe this job does not need these includes/macs/libs...

bash/runasmdemos
adelosa: Oh yeah. So when I ran this script it failed as there was not a .MLC version. I added the prefix so the job would run. Should the file be extension be renamed?

build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh -l
+set -e
+echo "::set-output name=javaversion::$(java -version)"
```
jyganci, line 3: What is the purpose of this line?

adelosa: See https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions. It's a github actions syntax to capture variables that can be used in scripts..

.github/actions/build/Dockerfile
```diff
@@ -0,0 +1,8 @@
+# Container image that runs your code
+FROM adoptopenjdk/openjdk15
```
adelosa: Question for @abekornelis @dsh33782 @jyganci - what version of Java do you want to use to build the package and test. As you can see, I am currently using AdoptOpenJDK v15. 
Options are 8 (LTS), 11 (LTS) or 16 (Latest version). I would suggest the latest LTS version (LTS=Long term support). See https://adoptopenjdk.net/. I don't suggest Oracle Java due to the current licensing issues.

jyganci:
I just looked through the source for each of the modules. Looks like they
do some COPY statements
for members in zcobol/lib. Since the assemblies automatically start their
sysmac, syscpy values with
'+', the directory containing the module being assembled is put in front of
it. That's why it works.

jyganci:
This is a question for Don. He's already pointed out that DEMOM8Q1.MLC was
missing.

Don,

In RUNASMDEMOS.BAT (and bash runasmdemos), is it okay to replace DEMOM8Q1
with DEMOM8Q1.ZSM?

.github/actions/build/Dockerfile
jyganci: I vote for Java 8. I use Oracle Java 1.8.0_201 which uses the older license and agree with Anthony that the current Oracle license should be avoided. Since the Java 8 update 8u221 on April 16, 2019, the Oracle JDK license changed to a more restrictive license. See the Oracle Java download page.

Hugh's comment about using "javac -source 8 -target 8 -encoding UTF-8" makes me wonder whether it should be incorporated into BLDJAR.BAT and corresponding bash script bldjar. This would allow a developer to use any free version of the JDK (Oracle JDK 8 prior to 8u221, any OpenJDK 8+ version) to build z390 for any JRE 8+ that a user might have.

build.sh
jyganci: Okay. Thanks.

abekornelis:
John, all,

I agree we should avoid the new Oracle license. Next steps are only too
predictable :-(

I also agree with the suggestion John made. I've been using a bldjar
thus modified for a long time.
I think those changes were lost due to that version never having been in
a distribution file.

No problem - these are really small changes and could be changed easily.

Kind regards,
Abe
===


Op 13/06/2021 om 16:03 schreef John Ganci:

bash/runasmdemos
abekornelis: I had a discussion about this program some time ago with Don. We then had both an mlc and a zsm version. Since the mlc was generated from the zsm we decided to keep only the latter. If it now needs to be renamed to mlc that's fine with me.

.github/actions/build/Dockerfile
abekornelis: Hugh is entirely right. I;m in favor of following his advice.

dsh33782 (dismissed):
Thanks for adding these tests including updated zopcheck to build regression tests.

adelosa:
I have created an initial build process for z390.
A release will automatically be generated when a tag is applied to a commit assuming all the tests run.
It is based on the linux build process.

Have a look at https://github.com/adelosa/z390/releases where you can find the releases notes and the zip file.
The body text is sourced from the CHANGELOG.md file which I have just created by renaming the README.TXT file that served the same purpose. I would suggest this is just a change log, not instructions for use.. thats what the docs are for.

Your are invited to try this zip file and let me know about any issues. Any questions, let me know. 
The next release of z390 will be using this process.

dsh33782:
I downloaded zip for v1.7.08 and successfully ran all the tests in bat, rt\bat, and rt\z390 GUI interface.
The only problem I encountered was trying to run bat\BLDJAR which failed with following messages:

W:\dsh6w\github\z390_v1.7.08>bat\bldjar
Could Not Find W:\dsh6w\github\z390_v1.7.08\jar\*.java
Could Not Find W:\dsh6w\github\z390_v1.7.08\jar\*.class
Could Not Find W:\dsh6w\github\z390_v1.7.08\jar\z390.jar
The system cannot find the path specified.
The system cannot find the path specified.
java version "1.8.0_261"
Java(TM) SE Runtime Environment (build 1.8.0_261-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.261-b12, mixed mode)
javac: file not found: *.java
Usage: javac <options> <source files>
use -help for a list of possible options
BLDJAR   ERROR: Error compiling z390

There is nothing in the jar directory and the original src directory is missing?

jyganci:
I was just about to report this issue.

The way the build process is working  running bldjar depends on the build
running in a git environment.
In particular, bldjar invokes relver, which creates a z390.properties file
that contains a "version" property using output from "git describe".

There should be no requirement on anyone downloading the zip file having
git installed. If "Joe User" downloads the zip file, the only requirement
for running z390 is having a JRE 1.8 or higher. If the user wants to build
the jar file, they must have a JDK 1.8+. There is no git requirement.

Regards,

John Ganci

bash/bldjar
```diff
@@ -1,53 +1,38 @@
...
 javac -d . -g:none ../src/*.java
 ec=$?
 echo "The exit code of the javac command is $ec"
-# pause to allow the user to terminate
-read -s -p "Press Enter to continue; press Ctrl-C to exit"
-echo ""
-echo ""
+
+# create z390.properties file
+../bash/relver
```
jyganci: It is my opinion that you cannot do this. See comments for relver.

bash/relver
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# relver: use git to set z390.properties version details
+
+set -e              # exit on all errors
+cd $(dirname $0)    # set to directory of script
+cd ..
+echo "version=$(git describe)" > z390.properties
```
jyganci: Line 8 requires that anyone running bldjar, which invokes relver, must (1) be using git, (2) be running bldjar from within the correct git repository. A version of z390 zip downloaded by "Joe User" only requires Java: (1) JRE 1.8+ if the user only intends to run z390; (2) JDK 1.8+ if the user intends to rebuild the z390 jar file (by running bldjar). There cannot (and should not) be a git requirement. Don's recent email illustrates the problem.

Don's bldjar attempt failed due to no src directory, not the relver issue. The src directory should be part of the zip file. Had the src directory been present, Don's bldjar would've succeeded compiling the source but the relver in bldjar would probably fail. See preceding paragraph.

adelosa:
Ok. I understand what you are saying. We are moving to a git based Dev model so some things will need to change. 
The Z390 distribution zip works without the source and git installed. What you cannot do is compile z390 Java from the distribution. 
My view is if you want to compile it yourself then you should clone the repo and do it that way. Distribution is for z390 users, not for z390 Java devs who should clone. 
Without a git dependency I cannot automate the build process.
If I add the source to the dist zip, it will compile but without a version number.
Let me know your thoughts.

.github/actions/build/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+FROM adoptopenjdk/openjdk8
+
+# install git so we can get version label
+RUN apt -y update;apt -y install git zip
+
+# Copies your code file from your action repository to the filesystem path `/` of the container
+COPY entrypoint.sh /entrypoint.sh
+
+# Code file to execute when the docker container starts up (`entrypoint.sh`)
+ENTRYPOINT ["/entrypoint.sh"]
```
jyganci, line 11: Installing git where?

For the remaining ".github/..." entries, I will view them but I do not have enough git expertise to check if they are valid.

.github/actions/build/entrypoint.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -l
+echo "Running ${GITHUB_WORKSPACE}/build.sh"
+${GITHUB_WORKSPACE}/build.sh
```
jyganci, line 3: Runs the build.sh bash script in the z390 root directory. I assume GITHUB_WORKSPACE is an environment variable set correctly.

.github/workflows/buildwin.yml
```diff
@@ -1,17 +1,14 @@
...
   build:
     runs-on: windows-2019
     steps:
       - name: Checkout main
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Build z390 (win)
         shell: cmd
         run: call .\BUILD.BAT
```
jyganci, line 14: I assume this is what ran in the problem I saw yesterday when I did the "Fetch upstream". Not sure why the windows version was run. Why not the Linux/Unix version? (Rhetorical question only.)

bash/blddist
```diff
@@ -0,0 +1,42 @@
...
+
+# add test and supplimental components
+supp_dirs=( assist demo guam qsam rt soa sort structuredmacros tests vsam1 vsam2 zopcheck zpar )
+for dir in ${supp_dirs[@]}; do
+    cp -R ${root_dir}/${dir} .        # unix bash scripts
+done
+
+# create zip package
+cd ..
+zip -r z390_${version_tag}.zip z390_${version_tag}
```
jyganci, line 42: Very nice.

I see that blddist is invoked from build.sh. No parm(s) passed to blddist. Is it invoked anywhere else?

I still object to the use of git commands in the build process (so far "git clean -dfX" and "git describe"), but do not see an easy way to avoid them.

The comments

bash/relver
jyganci: The last line completely replaces z390.properties. Right now it only has one line, so this is okay. However, if addtional properties are added at a later date, this will need to be changed.

I'd still like to find a non-git way to set the version property.

.github/actions/build/Dockerfile
adelosa: This file sets up a docker container to run the build. The container starts with a Java 8 image and I then install gut for the new build functions. This is all GitHub Actions stuff if you want to learn more.

.github/actions/build/entrypoint.sh
adelosa: This is set by GitHub Actions when running your build to point at the source.

.github/workflows/buildwin.yml
adelosa: So this branch has not been merged yet so that's why you did not see the Linux build. The windows one currently runs on push to main which is what happened in your repo when you fetched the upstream.

bash/runrtqsam
```diff
@@ -39,54 +31,54 @@ shopt -u nullglob
...
 
 export SYSUT2=qsam/TESTFTL.TFV
-${dir}/bash/asmlg qsam/BLDFTL.MLC trace $1
+bash/asmlg qsam/BLDFTL.MLC trace "TIME(30)" $1
 export SYSUT1=qsam/TESTFTL.TFV
-${dir}/bash/asmlg qsam/CHKFTL.MLC trace
+bash/asmlg qsam/CHKFTL.MLC trace
 
 ls qsam/TEST*.TFV
 echo "Verify runqsamrt"
```
jyganci: You might change "runqsamrt" to "runrtqsam".

bash/blddist
adelosa: Thanks. It's the power of Google 😁
I understand your reservations. 
If you have a better approach let me know and I will see what I can do.

bash/bldjar
adelosa: Alternatives requested.

bash/relver
adelosa: Yes for now. If someone comes up with other properties they want to include then we can change the process then. We also may never do that so didn't want to over engineer it. Just enough to have it working.

adelosa: A non git approach is possible. Let me know what you are thinking.

adelosa:
@dsh33782 mad a comment about Java versions. A Java 8 jar should run correctly and without change on all subsequent releases. Has anyone experienced this in real life?

rt/bat/RUNRTQSAM.BAT
```diff
@@ -48,7 +48,7 @@ CALL BAT\ASMLG QSAM\CHKFT.MLC TRACE || goto error
 SET SYSUT2=QSAM\TESTFTL.TFV
 CALL BAT\ASMLG QSAM\BLDFTL.MLC TRACE || goto error
 SET SYSUT1=QSAM\TESTFTL.TFV
-CALL BAT\ASMLG QSAM\CHKFTL.MLC TRACE || goto error
+CALL BAT\ASMLG QSAM\CHKFTL.MLC TRACE TIME(30) || goto error
 DIR QSAM\TEST*.TFV
 
 goto:eof
```
jyganci: You should probably add the ability to add NOTIME parameter to the BLD* steps. See bash/runrtqsam.

src/tz390.java
```diff
@@ -7684,4 +7688,27 @@ public boolean check_java_version(){
...
+            //get the property value and print it out
+            return prop.getProperty("version");
+
+        } catch (IOException ex) {
+            ex.printStackTrace();
+            return null;
+        }
+
+}
 }
```
jyganci, line 7724: This is a good way to get the version value. Think about finding a way to build z390.properties without using "git desribe".

bat/RELVER.BAT
```diff
@@ -0,0 +1,8 @@
+@echo off
+setlocal
+pushd %~dp0
+cd ..
+
+rem relver: use git to set z390.properties version details
+FOR /F "delims=" %%i IN ('git describe') DO set version=%%i
+echo version=%version% > z390.properties
```
jyganci: Same comments as for bash/relver.

jyganci (commented):
Finished reading through the 29 files. Some comments and suggestions added. Please read over them and respond if necessary.

bash/blddist
jyganci: I don't. Let's both think about it for a few days. If nothing jumps out, I'll approve this request so we can move on.

src/tz390.java
jyganci: You might want to modify the getVersion() method to add a finally clause to the try-catch so that the input stream is closed. This will require that you only have one return statement at the end. See

https://crunchify.com/java-properties-file-how-to-read-config-properties-values-in-java/

Here's one possibility for modified getVersion() based on the link. Line numbers are the above line numbers.
1.  Above 7693 (before the try) add
String result = null;

2. Replace 7697 -- 7706 with
if (inpt != null) {
    prop.load(input);
    result = prop.getProperty("version");
} else {
    System.out.println("Sorry, unable to find z390.properties.");
}

3. Delete 7710 (the "return null;" in the catch block)

4. Modify 7711 (the closing brace for the catch block) and add 2 lines:
} finally {
    input.close();
}

5. Add "return result;" above 7713.

The only problem with this (and the original getVersion()) is that if there is an error, the user may not see the console output ("Sorry, ..." and stacktrace) and where the version is missing in output may not be apparent. It might be better to set result to artificial values -- one if prop.getProperty returns null; one if no z390.properties file; one if I/O error.

bash/blddist
adelosa: Sounds like a plan.. I do like the versions being driven by tags and commits in the repo.. just makes it clear which version of the source the jar was built with - remember, its not just the version number but the number of commits and the hash if its a build between versions.
Each commit could look like this:  v1.7.07 -> v1.7.07-1-abcdef1 -> v1.7.07-2-1fedcba -> v1.7.08
Without git, how can you work this out? There may be a way, I just can't think of how without some artefact from the source control system.

bash/runrtqsam
adelosa: Fixed in 700b53e

rt/bat/RUNRTQSAM.BAT
adelosa: I have changed both bash and bat versions to accept parameter to bld runs. Also updated the build scripts to use TIME(30) due to the BLDFTL occasionally timing out on GitHub build. Have seen on my own workstation as well.
See 1d6b6cd for commit details.

adelosa: Oops.. didn't test windows.. put a $ instead of a % for variable.. Fixed in f63065e

src/tz390.java
adelosa: See deccb07 for changes made related to the above comments.

The try (resource) will automatically close the resource so no need for the finally.
see https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html for the syntax and https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html to confirm that InputStream object is autoclosable.
I am setting the version as "Uncontrolled" if the version cannot be found as null is not a good message.
I am also on the fence about displaying the messages -- the uncontrolled tells the story. Really need to get the message at compile time rather than run time.. will have a think about this.

src/tz390.java
```diff
@@ -7684,4 +7688,36 @@ public boolean check_java_version(){
...
+
+    } catch (IOException ex) {
+        System.out.println("Unable to set version - unexpected exception");
+        result =  "Uncontrolled";
+        ex.printStackTrace();
+    }
+    
+    return result;
+
+}
```
jyganci: It may be overkill, but it would help debugging if the setting of result to "Uncontrolled" on lines 7705, 7708, and 7716 were "Uncoltrolled-1", "Uncontrolled-2", and "Uncontrolled-3", respectively. Note that result is implicitly set to "Uncontrolled" at the "else" on line 7710, so the value of version would indicate exactly which error occurred. (The System.out.println output would probably be long gone by the time anyone noticed the error.)

I started to mention the missing "finally" block, but saw your response in email. Nice to know it is not needed.

jyganci:
Yes, a Java 8 jar should run correctly and without change on all subsequent
releases. However, I did have one (at least) instance where I had to
rebuild the jar that Don distributed so that it would run on my Raspberry
Pi, which has OpenJDK 11. I did the rebuild on the Pi.

Regards,

John Ganci

src/tz390.java
adelosa: No probs to link the version to the error. 

I have been working on enhancing the relver script so that if does the checking at compile time rather than run time. The check in java is runtime.. by the time you get here, it is too late.

Here is what I am thinking.. 
When called from bldjar, it would stop build if git not found, but provide a message to say you can ignore by providing the parameter force to bldjar. This would require passing $1 to the relver script, but thats it.
The dev would need to provide force and be told why a version is not provded and can continue to build without git installed. Version is set by script rather than Java.. so the errors above are more related to an incorrect build process.
Let me know what you think as an alternative.. 

``` bash
set +e                  # need to handle errors in this script
cd $(dirname $0)        # set to directory of script
cd ..

VER="Uncontrolled";

# is git installed
$(git --version > /dev/null 2>&1)
GIT_INSTALLED=$?
if [ ${GIT_INSTALLED} != 0 ]; then
    echo >&2 "Git is not installed";
else
    # is in a git repo
    $(git rev-parse --is-inside-work-tree > /dev/null 2>&1)
    GIT_REPO=$?
    if [ ${GIT_REPO} != 0 ]; then
        echo >&2 "Not a git repo.";
    else
        VER=$(git describe);
    fi    
fi

set -e

if [[ ${VER} == "Uncontrolled" ]]; then
    parm=$(echo "$1" | tr '[:lower:]' '[:upper:]');
    if [[ "$parm" != "FORCE" ]]; then
        echo "Version is uncontrolled - add FORCE parm to ignore"
        exit 8;
    fi
fi
echo "version=${VER}" > z390.properties
cat z390.properties;

```

dsh33782:
Anthony

 

Sounds ok to me.

 

Don Higgins

 ***@***.***> ***@***.***

 <http://www.don-higgins.net/> www.donhiggins.org

src/tz390.java
adelosa: The `bldjar` script now fails if run without git. You can still compile by running `bldjar force` but the build version will be set to "Uncontrolled". The reason for the error will be provided at this point. Both Win bat and Bash versions have been updated.
* Windows bc736e9
* Bash 2b4fb95

I have also added debugging in case someone does manage to compile and not use the bldjar script and one of the errors in TZ390 occur. If bldjar is used, then these errors should never happen.. see 3e45ca0

src/tz390.java
adelosa: Compile and run time checking has been added
* Windows bc736e9
* Bash 2b4fb95
* Unique versions for Java errors 3e45ca0

.github/workflows/buildwin.yml
adelosa: Oh.. so the fetch-depth:0 was added so that tags are included in the checkout which is required by the version logic. Otherwise tags are not retrieved in the checkout.

bash/bldjar
```diff
@@ -1,53 +1,38 @@
...
 ec=$?
 echo "The exit code of the javac command is $ec"
-# pause to allow the user to terminate
-read -s -p "Press Enter to continue; press Ctrl-C to exit"
-echo ""
-echo ""
+
+# create z390.properties file
+../bash/relver $1
+
```
jyganci, line 22: The bash/relver script must be run in a git environment in order to produce a valid z390.properties file.

It is still my opinion that bldjar should not require a git environment. A z390 user should not be required to have git in order to build z390.

Why can't the z390.properties file be built in a git environment, becoming a z390 file that already exists at the point bldjar is run? It should be included in the two "release" zip files.

bat/BLDJAR.BAT
```diff
@@ -20,8 +21,11 @@ if %ERRORLEVEL% NEQ 0 (
     echo BLDJAR   ERROR: Error compiling z390
     exit /b %ERRORLEVEL%
 )
+rem create z390.properties file
+CALL ..\BAT\RELVER.BAT %1
```
jyganci, line 25: See comments at bash/relver.

In short, the z390.properties file should exist before BLDJAR.BAT is run.

build.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -e
+echo "::set-output name=javaversion::$(java -version)"
+
+if [ "$1" == "clean" ];         # if "clean" parm passed
+then
+    git clean -dfX              # delete all git ignored files
+fi
+
```
jyganci, line 9: The build.sh script could be considered to be a "requires a git environment" script. Is it possible to invoke bash/relver at ths point? If so, then the bash/relver invocation in bash/bldjar could be removed, resulting in bash/bldjar no longer requiring git.

Might require making sure no subsequent "git clean" is done.

bash/bldjar
adelosa: Just to be clear, the properties file is embedded in the jar so is not required for run, only build. In this case, inclusion in the distribution is not required unless you need to build. And the dist is not for building, that's what the source is for.

bat/BLDJAR.BAT
adelosa: The version is dynamic so don't want to manage this statically in source. It's also a chicken and egg situation. To change the version you need to commit which changes the version etc. 
Also you can still build without Git using these changes but it is not version controlled and hence version label will say "Uncontrolled". An message will indicate no git and force Param can be used to continue the build. If you want a versioned build then you need a version control system. 
Hope this makes sense.

build.sh
adelosa: I can move the relver out of bldjar. Issue is bldjar could pick up a stale version if relver is not run immediately after a new commit.

bash/bldjar
adelosa: Also the build works without Git.. you need to use "force" param to ignore the error.. I have enhanced the build scripts to check for git.

jyganci: I know that the z390.properties is embedded in the z390.jar file and I know that tz390.java gets the version value at runtime by reading the embedded z390.properties file.

I also understand that, DURING DEVELOPMENT, the version value is in a state of flux so care must be taken with its use. However, when the "release" zip files ("build" and "source") are created, the value is frozen. In order to see the version value without running z390, one can extract the z390.properties file from the z390.jar file contained in the "build" zip file.

I will yield on requiring the z390.properties file being in the "build" zip file. However, it is my opinion that it should be a file in the "source" zip file.

There is no need to continue discussing this issue. Just do what you think is best.

adelosa: Hi @jyganci  
I do appreciate your feedback. I had a play yesterday to see if I could create a version that 

1. Includes the z390.properties file in the source
2. Does not link the bldjar script to relver.
3. bldjar has no reference to git

So here is what I came up with:
* A z390.properties file is added to root - Just has version=Uncontrolled and note not to update
* The build script calls relver to create **jar**/z390.properties (not in root, creates jar dir if does not exist)
* The build script calls bldjar
* bldjar checks for jar/z390.properties file. If it does not exist, copies from root to jar
* The jar is built as it does today

So this will remove any git dependencies from bldjar, and use the root z390.properties file.

If called from build script, then the jar/z390.properties file will be generated from relver and used by bldjar.

This can work. The question in my mind, is this better? Personally I like the way it is now because you always get a fresh version with every build, regardless of how the scripts are called (via build, or bldjar directly). This approach does leave the possibility of a stale version file. If we had more than the version property then it makes more sense. 

If you like this approach better, let me know and I will make the required changes. I'm easy either way at this point.

adelosa: @jyganci Hey John - If you are happy to leave this as is, can you approve the PR. If you want me to make the changes as per above, then let me know so I can update.

jyganci: Leave it as is. My concerns still stand, but we need to move on.

bash/blddist
jyganci: Where is the distributed "source" built?

jyganci (approved):
Concerns have been expressed, but we need to move on. Someone else can do the merge.

bash/blddist
adelosa: In the releases view, it always provides a link to a zipped up version of the repo for the specific release. 
See https://github.com/adelosa/z390/releases/tag/v1.7.08 for example.
You don't have to do anything explicitly for this to be created. Its just available.
image

src/tz390.java
adelosa: We are moving forward with using Git for versioning for lack of a better alternative.

PR #314, opened 2021-06-20
Windows automated build

### LINKED ISSUE #312: Automated build process - create the jar

adelosa:
This is effectively the same as PR #313, except run on a windows host.
Introduces the Windows version of build.sh in the root, the BUILD.BAT file.

If you have a JDK installed, you should be able to run BUILD.BAT from the root and run the build on your local machine.

BAT files are a bit different when it comes to dealing with errors. In shell, you can add `set -e`  and any error will stop processing but there does not seem to be anything similar in Windows BAT files. 

To support error checking many of the base scripts have been modified to added explicit checking of the %ERRORLEVEL% or use of the pipe command `|| goto error` with corresponding code to handle the error and return to the called.

Some other changes included here: 

1. Remove pause commands that don't play well with automation - add automated checks instead.
2. `setlocal` added so there is no environment details kept between/after script run
3. `pushd` used to cd into the script directory at start of the script. Allows script to return user back to directory they issued command at the end of script. Also allows the scripts to be run from any directory.
4. exit /b added to allow conditions to be cascaded back to the main script and handled.

**This has required changes to some of the core tool scripts - asm, asmlg etc.. so testing to make sure I have not broken them is advised.**

Some notes:

* I have only done changes where required to support the current BUILD.BAT process so some of the existing core scripts have not been changed but probably should be to align with the modified ones.
* It seems there is pauses designed for someone to do manual checks. These checks will need to be automated if used in the automated build process.


dsh33782 (dismissed):
I downloaded your pending changes and successfully ruan batch tests which no longer pause to review output.  I did run into the missing VCDT macro error in trying to run batch.bat.

dsh33782 (approved):
Anthony added missing VCDT macro. So I re-approving pull request.

PR #315, opened 2021-06-20
z390 docs structure

adelosa:
Another document related pull request focused on getting the content in the right structure and the adding of at least placeholders for all existing documentation content.

See preview of the new structure here:
https://adelosa.github.io/z390

* Home pages is a "What is z390?"
* Getting started provides simple instructions to run "hello world" type apps
* User guide will be the next step and go through the various features of the app as required by a user of z390
* Reference provides the detailed reference information.. 
* Contributing is where any docs related to developing z390 should be placed.

This is a work in progress, but I would like to get feedback on what content belongs where as well as suggested content and structure for the docs. 

**ALL SUGGESTIONS WELCOME**


dsh33782:
I really like the intro with guide to install both JDK and z390 and then how to code and run Hello World in both assembler and cobol.

abekornelis:
@adelosa would you please add the cobol compiler to the bullet list of features? It is mentioned at the top only ....
Also, would you please add the "Scripted Regression Testing Facility" ?

adelosa:
Hi @abekornelis. 
This page is the "Tell me why I should try z390!". 
Happy to add a COBOL reference.. Mainframe dev may want to create COBOL programs. 
Is it a specific variant of COBOL? Mainframe COBOL perhaps? Please suggest wording.. I had a hard time describing the COBOL compiler.
Not sure anyone comes to the site looking for a "Scripted Regression Testing Facility". If it was not there, but had the rest of the features, would people still try the product? Probably... More is less on this page.. 
I think this should be documented, but does not belong on the first "try me" page..
My views alone and happy to be overridden but please consider my approach..

dsh33782:
Anthony, I like the work you have been doing to move z390 docs to MD format with automated indexes as part of repository.  Being able to search all the docs will be most helpful in the future.

jyganci:
Couple of comments.

1. Getting started.
1.1. Download link doesn't work yet. Okay, since this is a first cut. When clicked, it should lead somewhere where the user can download the latest distribution zip file and the corresponding source.
1.2. Clicking the get Jave link leads one to AdoptOpenJDK site. Message there is that the site has moved. You might want to update this link.

2. Storage services.
2.1. Since this is really "SVC 4", why isn't this in the "SVC services" section?

dsh33782 (dismissed):
Draft looks ok to me

adelosa:
OK - I think the docs are "good enough" to merge. We can always improve as we move forward. I would like to see this included in the 1.8.1 release if we can.
I need to make a couple of changes to replace the current doco at https://z390development.github.io/z390 and point the download link to the latest version. Please hold off on your review until I complete these updates.

dsh33782 (dismissed):
okby me.

abekornelis (dismissed):
Looks good enough to me. Further improvements can be made on successive issues - let's try to apply maintenance to the doco in sync with modifying z390 behaviour.

dsh33782:
Abe, all

 

Good idea!  We should look for doc updates along with each issue update that changes or enhances z390 functionality.

 

Don Higgins

 ***@***.***> ***@***.***

www.don-higgins.net

dsh33782 (approved):
ok

abekornelis (approved):
Let's give it a go. From my perspective it looks good enough. 
Any issues that pop up can be addressed over time.

Issue #305, opened 2021-06-08
ASSIST XHEXI instruction does not behave as documented in the ASSIST User's Manual


jyganci:
When examining the z390 implementation of the ASSIST instruction XHEXI, it was noticed that it does not exactly follow the ASSIST User's Manual documentation.

The documentation follows.

>         XHEXI REGISTER,ADDRESS
>
>1. Beginning at the location ADDRESS, memory is scanned until the first non-blank character is found.
>2. If the first character found is anything bu a legal hexadecimal character (0-9,A-F), the condition code is set to 3 (overflow) and this address is placed in register 1. If REGISTER is anything but register 1, its contents remain unchanged.
>3. One to eight hexadecimal characters are scanned, the number converted to hexadecimal, and the result is placed in REGISTER. The value placed in the register is internal hexadecimal with leading zeros included and the number is right justified.
>4. Register 1 is set to the address of the first non-hexadecimal character. With this in mind, the user should not code register 1 as REGISTER. This allows you to scan across the card for any number of character strings. The strings should be separated by blanks. The end of the string could be flagged with any non-hexadecimal character and a test could be made after a Branch Overflow (see sample program).
>5. If more than eight hexadecimal characters are found, register 1 is set to the address of the ninth. This allows the user to scan across long strings of numbers.

Z390 HEXI violates 2. It does not set register 1 as indicated.

Z390 HEXI violates 3. (1) It scans hexadecimal characters until a non-hexadecimal character is found, so it could potentially scan more than eight hexadecimal characters. (2) It sets condition code 0.

Z390 HEXI violates 5 because it violates 3.

Comment: the ASSIST XHEXI is flawed because it does not set a condition code when a string is successfully converted. If the condition code were 3 before the XHEXI instruction and it successfully converts a string to hexadecimal, the condition code will still be 3, indicating that the first non-blank is not a hexadecimal character. Because of the flaw, a user must ensure that the condition code is not 3 before issuing XHEXI.

The z390 XHEXI could fix this flaw, but if it should behave as the documentation specifies, it should also have the flaw.

adelosa:
Fixed by PR #331. Closing

PR #331, opened 2021-11-20
ASSIST XDECI and XHEXI Is291 is305

jyganci:
Issue #291 and issue #305: (1) Fixes for the ASSIST instructions XDECI and XHEXI. (2) New assembler programs to test the changes. Note that the test programs accept a PARM value, lowercase v, that results in "verbose" output describing each test performed. If the PARM value is omitted, only summary statistics are printed. (3) Update bat/RUNASSIST.BAT and bash/runassist to invoke the test programs using "verbose".


adelosa (changes requested):
Hi @jyganci -- this is great and I love the test cases..  

I pulled down your branch, rebuilt the jar based on your code and then ran the bash/runassist script.
It all worked. 

I then I changed the expected value of one of the tests to something wrong and reran it. 
It detected the error, but the job was still successful (RC=0)

For the automated build to be useful, it needs to tell you when a test fails.
The test programs should return a non zero return code so that the build will stop with an error, otherwise we will never know it is broken.

I am not sure how to do this in an assembler program, but I am sure you will.
**edit** OK, so the value in R15 at exit is the RC value..

My other comments were around copyright.. I would like to do a bulk update on the copyright notices before our first 1.8 release so don't worry about that one.

jyganci:
Anthony,

Thanks for testing. Yes, the test programs always have a return code of
zero. I'd thought about that at one point, but decided that since the user
would see the error, there was no need. I was not thinking of running them
in automation.

It will be an easy fix to create a non-zero return code when an error
occurs. I'll try to get to it today (my today) but I want to finish up some
of the cics stuff I started.

Regards,

John Ganci

jyganci:
The changes have been made and pushed to my fork. TSTXDECI and TSTXHEXI now set the return code to 8 if there is at least one error. Tested each program with  "no errors" and "at least one error".

adelosa:
Looks good and happy to approve. I had a go at fixing myself for my own learning. I was on the right path but my instruction vocab is not that good. The register restore caused me the issues. I was at least at the right part of the program.

adelosa (approved):

PR #332, opened 2021-11-23
v1.7.8 release changelog

adelosa:
This pull request is to author the contents of the release CHANGELOG.MD file.

The CHANGELOG.md was previously the readme.txt file, which contained a changelog, project readme and some documentation
I have used a standard structure for this taken from https://keepachangelog.com/en/1.0.0/
I have removed any contents I would consider to be documentation - this should either go into the README.md file at the root of the project or be incorporated into the documentation in the doc folder.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,113 +1,32 @@
-Github z390 Branch V1707 has been created from the z390_v1707.zip distribution
-described below with the following additiona changes:
-1. z90.bat has been updated to use the start z390.jar command to release command line.
-2. The following additional directories have been retained from the Github prior
-   repository version v1706:
-   a. mfacc - sample assembler programs for student exercises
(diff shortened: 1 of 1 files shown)


dsh33782 (commented):
There are still problems with the zcobol cics transactions at the end cics\bat\dfhall.bat.
I'm working with Melvyn and John.  Today we have narrowed down the problem with cics\TESTBEC1.CBL to a problem in the zc390.java translator on 2 statements that have ended up generating 2 missing macros erroneously.   We could go ahead but it would be nice to get this fixed if we can figure it out in the next few days.

adelosa:
Hi @dsh33782. This pull request was just to come up with the change log. We will not release until we have all the things we want in the release. I have decided I am going to close this in pull request in favor of a new pull request that fixes issue we have with the alpha release and we can add the initial changelog into that one.

Issue #339, opened 2021-12-27
ZSTRMAC.MLC fix translator to convert structured macros to HLASM assembler compatible macros


dsh33782:
The current version of linklib\zstrmac.zsm does not correctly translate all structured macros in mad to HLASM compatible macros in mac\spm.  The last successful generation was back on v1506 in 2012 and several structured macro changes have occurred since them.  The conversion process requires assembling zstrmac.zsm with sysut1 set to input macro and sysut2 set to output macro using assembler option noasm.

dsh33782:
A work-around to generate the mac\spm version of z390 structured macros for use with HLASM has been found using zstrmac.mlc bootstrap version from z390 v1506.  To generate a macro set SYSUT1 to original z390 structured macro and SYSUT2 to unstructured HLASM compatible output macro.  Then assemble the zstrmac.mlc version which uses AREAD and PUNCH to generate the unstructured version.
The source for zstrmac.mlc and the bat file zstrmac.bat are attached are attached as txt files.
[ZSTRMAC.TXT](https://github.com/z390development/z390/files/7833539/ZSTRMAC.TXT)
[zstrmac1.TXT](https://github.com/z390development/z390/files/7833541/zstrmac1.TXT)
.

dsh33782:
When time permits, the zstrmac.zsm will be updated to handle the work-around and new structured macro changes currently in process for ACALL with parms

adelosa:
Pull request #342 has modified the build to generate the mac/spm folder in the distribution using rt/test/ZSTRMAC2.ZSM and remove it from the source. It does not use the attached text files above or linklib/ZSTRMAC.ZSM.
This fix should make sure that the ZSTRMAC build process uses the modified version of linklib/ZSTRMAC.ZSM in place of rt/test/ZSTRMAC2.ZSM. 
* bash/bldzstrmac and bat/BLDZSTRMAC.BAT generate the HLASM compatible SPM macros.
* bash/runzstrmactests and bat/RUNZSTRMACTESTS.BAT bootstrap and test the ZSTRMAC macros including the generated HLASM compatible version.

adelosa:
Just reviewing old issues. Is #84 the same issue or different?

PR #342, opened 2022-01-12
Structured programming macro/enhancements test and gen

adelosa:
This pull looks to test and generate components related to the SPM/SPE facility.

Currently the mac/spm folder has a static version of the SPM macros that have been generated to remove SPE. The Structured Programming Macros located in the mac folder are the gold copy.

A single README.md file has replaced the contents of the mac/spm folder with details on how to regenerate the macros if required if using source. The distribution has been updated to add the generated mac/spm folder.
So no change for distribution users, source users will need to run script to get the files.

2 new scripts (bash and bat) have been added:
* runzstrmactests - test the spm macros including after removal of SPE
* bldzstrmac - generate the non SPE version of the macros at the folder specified.

2 scripts have been modified (bash and bat)
* build - main build script now runs the tests in runzstrmactests
* blddist - the distribution has been updated to include a non SPE version of the mac's in mac/spm

Some outstanding questions:
1. It seems that rt/test/ZSTRMAC2.ZSM works for this whereas linklib/ZSTRMAC.ZSM does not. Should we replace the broken linklib version with the rt/test version and remove from rt/test?
2. Are the tests in rt/test/TESTSPM1.MLC sufficient to test all the currently supported features of SPM or does this need to be updated? 
3. I still need to look for any stray references to generating the macros.. if you notice any, let me know.


dsh33782 (approved):
These changes look good to me to support generation of  mac\spm HLASM compatible versions of the existing z390 structured macros such as AIF AELSE AEND.  In future release zstrmac translator will need to be updated to support updated ACALL APARM and AENTRY structured macros.

dsh33782:
Regardng questions:
1.  Yes, I recommend moving rt version of zstrmac to linklib and removing it from rt.
2.  Yes, there will need to be additional translation changes and regression test changes to support updated SPM macros still in development including ACALL APARM and AENTRY.
3. I don't think there are any other references to the existing mac\spm generated macros, and this is a great step to replace the static version of spm macros.

adelosa:
Thanks for the approval @dsh33782 - Rather than update here, I have added some comments to #339 so that we do the cleanup of the ZSTRMAC.ZSM location. This issue already covers the ACALL, APARM and AENTRY updates.

Issue #195, opened 2021-04-08
RPI 1622: Issue S0D3 when executing TRAP2/TRAP4 instructions. Not supported.


adelosa:
Raised by Dave Kienow 2018/08/01
Issue S0D3 when executing TRAP2/TRAP4 instructions; S0C2 for as yet unsupported privileged instructions: CSCH, CSP, CSPG, DIAGNOSE, ESEA, HSCH, IDTE, IPTE, ISKE, LASP, LCTL, LCTLG, LPSW, LPSWE, LPTEA, LRA, LRAG, LRAY, LURA, LURAG, MSCH, PALB, PCKMO, PFMF, PGIN, PGOUY, PTAB, PTF, RCHP, RRBE, RRBM, RSCH, SAL, SCHM, SCK, SCKC, SCKM, SCKPF, SIGP, SPT, SPX, SSCH, SSKE, SSM, SSTPS, STAP, STCKC, STCRW, STCTG, STCTL, STFL, STIDP, STNSM, STOSM, STPT, STPX, STRAG, STSCH, STSI, STURA, STURG, TB, TPI, TPROT, TRACE, TRACG, TSCH, XSCH

PR #351, opened 2022-02-10
Fix for Issue #195 (=RPI 1622) Issue S0C2 if any of the privileged instructions are used

abekornelis:
S0D3 for TRAP2/TRAP4


abekornelis:
Instruction STIDP is privileged according to PoP, now throws a S0C2 abend, but before it was implemented as an executable instruction storing 'z390' as processor ID. We cannot have our cookie and eat it - so now testins2 in directory tests is broken.
Either we implement STIDP as a non-privileged instruction, *or* we go with compatability and break all programs that use the STIDP instruction's incompatible implementation. I'd go for compatability with the PoP....

dsh33782:
Abe, all

 

I agree we should follow POP whenever possible and prevent supervisor instructions from executing.

 

You also pointed out another problem we should fix in v1.8.1 if not before.  There are duplicate MLC files in z390\tests and z390\rt\test including TESTINS2.  The one in rt\test fails with S0C2, but the one in tests runs ok due to fix I added in 2019 which was the only fix since 2012.

 

I’m working on Excel spreadsheet documenting 238 rt\test programs most of which are not currently being tested by BUILD.BAT.

I’m identify which ones are currently included and which ones I think should be added in priority order.

 

Don Higgins

 ***@***.***> ***@***.***

www.don-higgins.net

src/pz390.java
```diff
@@ -7246,21 +7305,27 @@ private void ins_B9XX(){
...
-				ins_setup_rre();
-				break;	
 		case 0xAA: // 5250 "B9AA" "LPTEA" "RRF-b" Z9-19
+			set_psw_check(psw_pic_priv); // RPI 1622 DK Privileged operation exception
 			ins_setup_rrf3(); // dsh rpi 2202
 			break;
+		case 0xAC:  // "B9AC=IRBM,14,144", // B9AC RRE 14,144 IRBM R1,R2 RPI 2202
+			set_psw_check(psw_pic_priv); // RPI 1622 DK Privileged operation exception
+				ins_setup_rre();
+				break;
```
jyganci: Lines 7321, 7322 appear to be indented too much.

src/pz390.java
```diff
@@ -2788,9 +2790,11 @@ private void ins_lt_c0() {
 		 */
 		switch (opcode1) {
 		case 0x80: // 1530 "8000" "SSM" "S"
+			set_psw_check(psw_pic_priv); // RPI 1622 DK Privileged operation exception
 			ins_setup_s();
 			break;
```
jyganci: Shouldn't the code check to see if in supervisor state (PSW bit 15 is 0) before the "set_psw_check(psw_pic_priv);" ?

Yes, today's pz390.java doesn't properly build the PSW to reflect problem/supervisor state. It should. I'm willing to look into this. It should be a simple change -- always set PSW bit 15 to 1 (problem state).

Suppose a user adds a new SVC to sz390.java that, when invoked, switches to supervisor state. (Their own MODESET macro that anyone can successfully invoke.) At that point, they should be able to use the privileged instructions.

Of course, this check would be needed at all the other places below where psw_pic_priv is set.

src/pz390.java
```diff
@@ -4671,9 +4720,11 @@ private void ins_B2XX(){
 			ins_setup_ie();
 			break;		
 		case 0xFC: // "B2FC S 7,72 TABORT D2(B2) z15
+			set_psw_check(psw_pic_priv); // RPI 1622 DK Privileged operation exception
 			ins_setup_s();
 			break;		
 		case 0xFF: // 3400 "B2FF" "TRAP4" "S"
+			set_psw_check(psw_pic_spec_op); // RPI 1622 DK Special Operation Exception
 			ins_setup_s();
 			break;
```
jyganci: Principles of Operation (SA22-7832-12), p10-180, states that

'When the PSW control in the trap control block is zero, the operation is as described above except as follows:
- Bit 31 of the current PSW must be zero; otherwise, a special-operation exception is recognized. ..."

Shouldn't these checks be here, before the "set_psw_check(psw_pic_spec_op);" ?

jyganci:
Comments added to #351 in Github.

src/pz390.java
abekornelis: Technically, you are right. Consequentially, we should then also change these opcodes from no-ops to really operational opcodes in case the PSW is on. That's quite a challenge. I would be okay with adding the supervisor state check but will not build all code required to make all privileged instructions run as described in the PoP. Quite a few of those cannot even be implemented in the z390 environment. Because we have no address spaces, page tables, etc.
If we add the supervisor bit check but leave the instructions as a no-op for the supervisor state case, would that be good enough for the time being?

src/pz390.java
abekornelis: The operation is defined as getting the trap control block address from the DUCT. We have no DUCT implemented in z390. We could do, but that would be a major change. For now, changing the operation from a no-op to a S0C2 (at least in problem mode) is an improvement. Further improvements could be made, but I'd rather see a separate story for such a change.

src/pz390.java
jyganci: Suggestion.
1. Create new submodule, say verify_supervisor_state. Could be something like
```
void verify_supervisor_state() {
    // TODO: add check of PSW to determine whether supervisor state
    set_psw_check(psw_pic_priv);
}
```
2. Replace all the current set_psw_check(psw_pic_priv) invocations with verify_supervisor_state().
3. At this point, the code is logically equivalent to what exists now.
4. At some future date, implement the TODO.

This greatly reduces the burden on whoever maintains the code.

src/pz390.java
jyganci: At least add a comment. Something like "// See PoOp for TRAP4. Additional checks needed."

src/pz390.java
abekornelis: Good suggestion! I'll give it a try later today or tomorrow...

src/pz390.java
abekornelis: Also a good suggestion. I'll look up every instruction affected, then add some relevant comments. That should help on the next round of changes. thanks John!

dsh33782:
All

 

I left out the text for supervisor state to save instructions in the emulator and leave it up to individual instruction code to do something or set pgm_chk flag.  The supervisor instructions which are a minority, could chk for supervisor state.

 

Don Higgins

 ***@***.***> ***@***.***

www.don-higgins.net

jyganci:
All,

If it were up to me, I'd probably add a new boolean variable that
represents the state -- true (1) = problem state, false (0) = supervisor
state. Whenever the PSW needs to be displayed, the new variable would be
used to set the corresponding displayable PSW bit accordingly.

Since z390 supports many different PSW formats (BC mode, EC mode, etc),
this will require nailing down
-  exactly which formats are supported
-  which format is in effect at runtime.

Comments? Suggestions?

Regards,

John Ganci

src/pz390.java
```diff
@@ -2788,10 +2798,16 @@ private void ins_lt_c0() {
 		 */
 		switch (opcode1) {
 		case 0x80: // 1530 "8000" "SSM" "S"
+             if (psw_problem_state == psw_problem_mode)
+                {set_psw_check(psw_pic_priv); // RPI 1622 DK Privileged operation exception
+                 } // in problem state issue privileged-operation exception, otherwise no-op until we implement this instruction
```
jyganci, line 2803: Code looks okay. Suggestion. Since psw_problem_state is a boolean, you could replace the if statement with
```
if (psw_problem_state)
```

src/pz390.java
```diff
@@ -2788,10 +2798,16 @@ private void ins_lt_c0() {
...
 			ins_setup_s();
 			break;
 		case 0x82: // 1540 "8200" "LPSW" "S"
 			psw_check = false;
+            if (psw_problem_state == psw_problem_mode)
+               {set_psw_check(psw_pic_priv); // RPI 1622 DK Privileged operation exception
+                } // in problem state issue privileged-operation exception
```
jyganci, line 2810: Shift these 3 lines right one position. There are more instances below.

src/pz390.java
abekornelis: I understand your suggestion. However, I do prefer an explicit test. To me, that is more readable and it lowers the risk of confusion as I don't have to remember the semantics of 'true' versus 'false'.

src/pz390.java
abekornelis, line 2815: layout is not really consistent. I agree it could be better. Current code indents to either column 14 or 15, varying throughout the entire source module. See e.g. SSM which has both indents :-( 
I'm in favor of running all source modules through a formatter to make it all consistent. But I'd prefer to do it as a separate change, not as part of this one...

jyganci:
Abe, all

Understood for both the "if" statement and the code indention. I've deleted
the 2 comments.

Regards,

John Ganci

jyganci:
I stopped my review of this pull request so that I could better understand how it is implemented. The remainder of the comments here pertain to testing done using the SSM and LPSW instruction implementations in pz390.java, implemented at lines 2790, respectively 2791, of the current version of pz390.java in the z390 repository. First the code, then the comments. The code is in pz390 submodule ins_lt_c0(), called from the execution mainline. Before the call, psw_check is set to true.

1. SSM. Lines 2790 -- 2792.
```
		case 0x80: // 1530 "8000" "SSM" "S"
			ins_setup_s();
			break;
```
2. LPSW. Lines 2793 -- 2797.
```
		case 0x82: // 1540 "8200" "LPSW" "S"
			psw_check = false;
			ins_setup_s();
			set_psw_loc(mem.getInt(bd2_loc + 4));
			break;
```
3. SSM comments.
Note that only instruction setup is done. When the submodule returns, psw_check is still true. The mainline code checks the value of psw_check. Because it is true, ABEND S0C1 is issued. The ins_setup_s() call in item 1 updated the next instruction address to the instruction following the SSM. A test program, TSSM.MLC, was written and run to verify that the S0C1 occurs and that the address in the PSW is the address of the instruction following the SSM.

4. LPSW comments.
Note that the code in item 2 sets psw_check to false, calls ins_setup_s(), then executes the LPSW instruction!!! What this means is that a problem state program can execute LPSW on z390. A test program, TLPSW.MLC, was written and run to verify that no abend occurs, the LPSW is executed, (ends up as a NO-OP here because the address is just past the LPSW, skipping over 2 bytes of zeros).

When the code in 2 returns to the mainline, psw_check is false, so the mainline code returns to the top of its loop to fetch the next instruction.

5. LPSW code with first cut at changes for invoking ABEND S0C2.
```
		case 0x82: // 1540 "8200" "LPSW" "S"
			psw_check = false;
                        if (psw_problem_state == psw_problem_mode)
                        {
                            set_psw_check(psw_pic_priv);
                        }
			ins_setup_s();
			set_psw_loc(mem.getInt(bd2_loc + 4));
			break;
```
6. At first glance, the 4 added lines seem to be what is needed. However, that is not the case. An existing pz390.java was modified to add just 1 line instead of the 4 lines:
```
		case 0x82: // 1540 "8200" "LPSW" "S"
			psw_check = false;
                        set_psw_check(psw_pic_priv);
			ins_setup_s();
			set_psw_loc(mem.getInt(bd2_loc + 4));
			break;
```
When TLPSW.MLC was run, ABEND S0C2 did occur. However, the PSW at abend contained the address of the instruction preceding the LPSW instruction, not the expected address of the instruction following the LPSW instruction. This happened because the ins_setup_s() had not been called yet (the ABEND S0C2 is generated in the set_psw_check() call sequence), so the address of the next instruction was still the LPSW instruction and the current instruction length was still the length of the previous instruction (for TLPSW, this is "SR    0,0"), resulting in the invalid  PSW at ABEND value. This was verified by running TLPSW.MLC.

7. SSM code with first cut at changes for invoking ABEND S0C2.
```
		case 0x80: // 1530 "8000" "SSM" "S"
                        if (psw_problem_state == psw_problem_mode)
                        {
                            set_psw_check(psw_pic_priv);
                        }
			ins_setup_s();
			break;
```
8. When TSSM.MLC was run with a modified pz390.java, the behavior was as in item 6. That is, the ABEND S0C2 occurred, but the PSW at ABEND contained the address of the instruction preceding the SSM instruction, not the expected address of the instruction following the SSM instruction. Reason for the behavior is the same as for item 6.

9. The TLPSW program was run on Hercules MVS 3.8. Its behavior was as expected. That is, ABEND S0C2 and PSW at ABEND contained the address of the instruction following the LPSW instruction.

10. On Hercules MVS 3.8 the TSSM program generated ABEND S0C1. The SSM instruction is not a valid instruction there.

11.  The two test programs, TLPSW.MLC and TSSM.MLC are attached for your perusal.
[TLPSW.MLC.txt](https://github.com/z390development/z390/files/8084311/TLPSW.MLC.txt)
[TSSM.MLC.txt](https://github.com/z390development/z390/files/8084312/TSSM.MLC.txt)

The problems shown above could be overcome by moving the check for problem state after the instruction setup code. That way, the address of the next instruction and the length of the current instruction will be set correctly.

Issues.
1. There are some instructions that can program check "early" so that the PSW at abend is the offending instruction, not the next instruction. More research should be done to see if any S0C2 candidates fall into this category.
2. The code that checks for problem state must be thought through again.
3. There are other considerations. For example, the call to set_psw_check(psw_pic_priv) results in a call to psw_pic_handler(), which calls sz390_svc_abend() to produce the abend. However, that call is preceded by checks for ESPIEs and ESTAEs, which may be invoked if there are any ESPIEs or ESTAEs active. If one of ESPIE or ESTAE gets control, the ABEND S0C2 is not done. The psw_pic_handler() code returns to set_psw_check() which returns to the SSM or LPSW code. The logic here must be checked carefully to make sure there's not a way for a problem state program to somehow switch to supervisor state.

I will continue to review the pull request, but I think it needs to be set aside until the problems and issues listed above are resolved.

adelosa:
Closed in error - reopening.

rt/mlc/TLPSW.MLC
```diff
@@ -0,0 +1,198 @@
...
+         MACRO
+         XDUMP
+         STMG  0,15,XREGS          Save registers
+         IPM   0                   Get CC and program mask
+         ST    0,XCCPM             Save CC and program mask
+         BAS   14,DUMPREGS         Dump the registers and CC, pgm mask
+         L     0,XCCPM             Saved CC and program mask
+         SPM   0                   Restore CC and program mask
+         LMG   0,15,XREGS          Restore registers
+         MEND
```
jyganci: The macro is not used here. Can be deleted. More lines at bottom of file.

rt/mlc/TLPSW.MLC
```diff
@@ -0,0 +1,198 @@
...
+*
+         DS    0D
+         DS    F
+XCCPM    DS    XL4                 Condition code and program mask
+XREGS    DS    XL128,Xl1           Registers: 0--15
+*
+DW       DS    D,XL1               Doubleword work and pad
+FW       DS    F,XL1               Fullword work and pad
+H2P      EQU   *-240
+         DC    C'0123456789ABCDEF'
```
jyganci: Part 2 of 2 of lines that are not used here. They can be deleted. The macro at the top and these lines were created for debugging purposes. Used in other test programs but not this one.

src/sz390.java
```diff
@@ -2720,9 +2722,9 @@ private String dump_psw(){ // RPI 819
 	 *   1. bits 0-7 x'07' translation, 
 	 *      I/O interrupts, and external
 	 *      interrupts enabled.
-	 *   2. Bits 8-15 x'85' key eight, machine                        // RPI 1544
-	 *      checks enabled, and problem state
-	 *      enabled.
+	 *   2. Bits 8-15 x'85' key eight, machine checks enabled                      // RPI 1544
+	 *      and problem state enabled.
+     *      Bit 15: Problem state (should be on: no SVC mode in z390)              // #195
```
jyganci: Suggestion.

First, reword item 2 as "Bits 6-14 x'84' key eight, machine-check enabled."

Second, add new item 3: "3. Bit  15  1   problem state enabled. (No supervisor state in z390.)"

Third, existing items 3-6 become 4-7.

src/sz390.java
```diff
@@ -2735,8 +2737,9 @@ private String dump_psw(){ // RPI 819
      * return 32 character hex PSW in format as definde by z-POP:     // RPI 1506
      *   1. bits 0-7 x'07' PER disabled, DAT enabled                  // RPI 1506
      *      I/O and external interrupts enabled                       // RPI 1506
-     *   2. Bits 8-15 x'85' key eight, machine check enabled          // RPI 1506 RPI 1544
-     *      No Wait state, Problem state                              // RPI 1506
+     *   2. Bits 8-15 x'85' key eight, machine checks enabled         // RPI 1506 RPI 1544 #195
+     *      No Wait state                                             // RPI 1506          #195
+     *      Bit 15: Problem state (should be on: no SVC mode in z390) // RPI 1506          #195
```
jyganci: Same suggestion as above.

src/sz390.java
```diff
@@ -2779,6 +2784,7 @@ private String dump_psw16(){                                          // RPI 200
      *      I/O and external interrupts enabled                       // RPI 2008
      *   2. Bits 8-15 x'85' key eight, machine check enabled,         // RPI 2008
      *      not wait state, problem state                             // RPI 2008
+     *      Bit 15: Problem state (should be on: no SVC mode in z390) // RPI #195
```
jyganci: Same suggestion as for dump_psw(). Wording for item 2 is different here. Change to "Bits 8-14 x'84' key eight, machine-check enabled, not wait state." Still need new item 3 as in dump_psw() and renumber current 3-6 to 4-7.

src/sz390.java
```diff
@@ -5958,7 +5966,12 @@ private void exec_test_cmd(){
                 String mask  = tz390.get_hex(pz390.psw_pgm_mask, 1);                          // RPI 1507
                 String mask_descr;                                                            // RPI 1507
                 tz390.put_trace(" PSW  "+dump_psw());                                         // RPI 1507
-                tz390.put_trace("      KEY   8     MODE: Problem");                           // RPI 1514
+                if (pz390.psw_problem_state == psw_problem_mode)                              //          #195
+                   {tz390.put_trace("      KEY   8     MODE: Problem");                       // RPI 1514 #195
+                    }                                                                         //          #195
+                else                                                                          //          #195
+                   {tz390.put_trace("      KEY   8     MODE: Supervisor");                    //          #195
+                    }                                                                         //          #195
```
jyganci: Nothing wrong. Just a suggestion. Could also do
```
String prob_descr = ((pz390.psw_problem_state == psw_problem_mode) ? "Problem" : "Supervisor");
tz390.put_trace("      KEY   8     MODE: "+prob_descr);
```
WARNING: haven't tested a compile of this!

abekornelis: As you can see I am not a real Java Jockey ... ;-)

src/pz390.java
```diff
@@ -2788,14 +2798,24 @@ private void ins_lt_c0() {
 		 */
 		switch (opcode1) {
 		case 0x80: // 1530 "8000" "SSM" "S"
+			psw_check = false;
```
jyganci: Since the SSM instruction does not have an implementation, you should delete this line, leaving psw_check set to true. That way, a supervisor state program would get an S0C1.

abekornelis: I will leave this as is - it is a standard semantic unit for all instructions in pz390. In supervisor mode, I will turn it back on with a comment to replace that statement with payload code for implementing the instruction.

src/pz390.java
```diff
@@ -4229,50 +4349,136 @@ private void ins_B2XX(){
...
-			break;
+            psw_check = false;
+            ins_setup_s();
+            if (psw_problem_state == psw_problem_mode)
+               {set_psw_check(psw_pic_priv); // RPI 1622 DK Privileged operation exception
+                } // in problem state issue privileged-operation exception, otherwise no-op as we have no address spaces
+            else // supervisor mode applies
+               {psw_check = true; // force S0C1 until we have a proper implementation
+                }
+            break;
```
jyganci, line 4466: SAC is used by any program to set Access Register Mode. The privileged operation only applies for an attempt to set the home=space mode in the problem state.

src/pz390.java
```diff
@@ -4666,16 +5140,27 @@ private void ins_B2XX(){
...
+            psw_check = false;
+            ins_setup_s();
+			set_psw_check(psw_pic_spec); // RPI 1622 DK Specification exception
+            break;
 		case 0xFF: // 3400 "B2FF" "TRAP4" "S"
-			ins_setup_s();
-			break;
+            psw_check = false;
+            ins_setup_s();
+            set_psw_check(psw_pic_spec); // RPI 1622 DK Special operation exception as we have no DUCT
```
jyganci: Should psw_pic_spec be psw_pic_spec_op?

src/pz390.java
```diff
@@ -3924,8 +4030,10 @@ else if (psw_amode_bit == psw_amode24_bit)               // RPI 2004
 			set_psw_amode(psw_amode31_bit | psw_extended_amode64_bit); // RPI 1506
 			break;
 		case 0xFF: // 80 "01FF" "TRAP2" "E"
-			ins_setup_e();
-			break;
+            psw_check = false;
+            ins_setup_e();
+            set_psw_check(psw_pic_spec); // RPI 1622 DK Special operation exception as we have no DUCT
```
jyganci: Should psw_pic_spec be psw_pic_spec_op?

src/pz390.java
abekornelis: That is completely correct. However, z390 does supports neither virtual memory nor even address spaces. The concept of AR mode, home space mode, etc. are undefined in the z390 context. Therefore any attempt to use SAC should result in an abend to warn the programmer that the program cannot execute as intended. For now I have chosen to follow the 'standard' resolution implemented for most privileged instructions. To my mind that should be good enough for the time being.

jyganci: No argument about issuing an exception. However, it is my opinion that the standard z390 technique of S0C1 via
```
case <instruction>:
    ins_setup_xxx();
    break;
```
is more appropriate for SAC. I'll defer to your judgment on how to do it.

jyganci (approved):

PR #360, opened 2022-03-04
#359 Fix tron handling in bat files

### LINKED ISSUE #359: bat scripts failing to run correctly in tron mode
Quite a few bat files in the \bat directory have an option to specify TRON as a first argument, which is intended to trigger trace mode for the script. Unfortunately, some recent changes cause the script to behave differently when trace mode is active.
The code snippet used in the intro of the script looks like this:
`if %1 == tron ECHO ON`
`if %1 == tron shift`
`if %1 == TRON ECHO ON`
`if %1 == TRON shift`

This snippet has the following shortcomings:
1. fails when %1 is not present
2. does not recognize mixed-case versions of TRON
3. the shift replaces %0 which causes the aforementioned problems 

Suggested solution is to replace the code snippet above in all bat files where they occur with the following:
`if %1.   == .    (echo %0 ERROR: missing file name`
`                  exit /b 16`
`                  )`
`if /I %1 == tron (ECHO ON`
`                  shift /1`
`                  )`

The problem occurs in only 9 bat scripts:
- bat\ASM.BAT
- bat\ASML.BAT
- bat\ASMLG.BAT
- bat\CBLC.BAT
- bat\CBLCL.BAT
- bat\CBLCLG.BAT
- bat\EXEC.BAT
- bat\LINK.BAT
- bat\MAC.BAT

abekornelis:
Fixing Issue #359 bat scripts failing to run correctly in tron mode

### DIFF
--- a/bat/ASM.BAT
+++ b/bat/ASM.BAT
@@ -1,10 +1,13 @@
 @echo off
 setlocal
 rem asm run macro assembly to generate relocatible obj from mlc source 
-if %1 == tron ECHO ON
-if %1 == tron shift
-if %1 == TRON ECHO ON
-if %1 == TRON shift
+if %1.   == .    (echo %0 ERROR: missing file name
+                  exit /b 16
+                  )
+if /I %1 == tron (echo on
+                  shift /1
+                  )
+
 if exist %1.BAL erase %1.BAL
 if exist %1.PRN erase %1.PRN
 if exist %1.OBJ erase %1.OBJ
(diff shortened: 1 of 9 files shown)


dsh33782 (dismissed):
Looks ok to me.

jyganci (dismissed):
Looks good. I assume each has been tested with 1) no parms 2) parm 1 = tron 3) parm 1 other than tron.

jyganci (approved):
Looked good before. Even more bulletproof now.

PR #361, opened 2022-03-04
Issue #357 Ending apostrophe on MNOTE not recognized

### LINKED ISSUE #357: Ending apostrophe on MNOTE not recognized
In macro mac\ZCLOSE.MAC on line 62 there's a MNOTE 12 statement:
`                   AIF   ('&OPTION' NE '')
                         MNOTE 12,'CLOSE OPTION NOT SUPPORTED - &OPTIONX
               '
                         MEXIT
                   AEND
`
When these lines are executed during assembly the terminating apostrophe on the continued line is not recognized and the following error is issued:
`MZ390E error 261     (8/116)116 MNOTE/PUNCH text must be in single quotes12,'CLOSE OPTION NOT SUPPORTED - &OPTION          MEXIT`

It appears the continuation is not handled correctly. Most noteworthy is the fact that the MEXIT on line 64 is included in the error message, which is absolutely unwarranted.

Please investigate what causes the continuation handling to fail and fix the underlying java code.
Changing the macro code could resolve this issue too. But then the underlying bug would remain and might recur on any user's assembly if it encounters comparable macro code.

abekornelis:
Issue #357 Ending apostrophe on MNOTE not recognized


adelosa:
Fixes #357

jyganci (approved):
Good catch!

PR #364, opened 2022-03-08
Fix issue #363 Various bat files are failing to return error code when processing has failed

### LINKED ISSUE #363: Various bat files are not returning an error code when processing has failed
Some bat files end their processing with a block like this:
`if %ERRORLEVEL% NEQ 0 (`
`    set RC=%ERRORLEVEL%`
`    erase %1.OBJ`
`    echo ASM      ERROR: See errors on mz390 generated bal file and console`
`    exit /b %RC%`
`)`

The problem is that the if clause is evaluated in its entirety before it is being processed.
During the evaluation/substitution phase the value of RC is not yet known.
Hence, the script will always terminate with an errorlevel/returncode of 0.

the issue can be fixed by splitting the if clause into an assignment part and a separate exit part.

abekornelis:
For now, this is a partial solution in that not all bat files were modified to support tron mode.
The change currently addresses only the build procedure and all its called files.
Solve Issues #363 and #370


bat/ASM.BAT
```diff
@@ -1,28 +1,41 @@
-@echo off
+@if /I "%1" == "tron" (echo on) else (echo off)
```
jyganci, line 1: What happens if user runs "bat/ASM.BAT" with no operands?

bat/ASM.BAT
```diff
@@ -1,28 +1,41 @@
...
-    erase %1.OBJ
-    echo ASM      ERROR: See errors on mz390 generated bal file and console
-    exit /b %RC%
-)
+call %~dps0mz390 %z_TraceMode% %1 sysmac(+%z_HomeDir%mac) syscpy(+%z_HomeDir%mac) %2 %3 %4 %5 %6 %7 %8 %9
+set z_ReturnCode=%ERRORLEVEL%
+if %z_ReturnCode% GTR 4 (erase %1.OBJ
+                         echo %0 ERROR: See errors on mz390 generated bal file and console
+                         goto return
+                         )
```
jyganci: Current code only tests for not zero. Has the new code been tested with programs that result in %ERRORLEVEL% values that are not zero?

bat/ASM.BAT
abekornelis: you will get the error message define in lines 14-17

bat/ASM.BAT
abekornelis: I changed that on purpose as ERRORLEVEL 4 indicates a warning only.
I will re-run the tests to double-check 0, 4, 8 are handled correctly

bat/ASMLG.BAT
```diff
@@ -19,21 +27,31 @@ if exist %1.STA erase %1.STA
...
-    exit /b %ERRORLEVEL%
-)
+call %~dps0ez390 %z_TraceMode% %1 %2 %3 %4 %5 %6 %7 %8 %9
+set z_ReturnCode=%ERRORLEVEL%
+if %z_ReturnCode% GTR 4 (echo %0 ERROR: see errors on ez390 generated log file and console
+                         goto return
+                         )
+if %z_ReturnCode% NEQ 0 (echo %0 WARNING: see messages on ez390 generated log file and console
+                         goto return
+                         )
```
jyganci: ASMLG.BAT has no control over what the ez390 return code value means. Also, the above call to EZ390.BAT only checks for return code not zero, in which case it echos the above "ERROR" message.

bat/ASM.BAT
```diff
@@ -1,28 +1,41 @@
...
-if %ERRORLEVEL% NEQ 0 (
-    set RC=%ERRORLEVEL%
-    erase %1.OBJ
-    echo ASM      ERROR: See errors on mz390 generated bal file and console
-    exit /b %RC%
-)
+call %~dps0mz390 %z_TraceMode% %1 sysmac(+%z_HomeDir%mac) syscpy(+%z_HomeDir%mac) %2 %3 %4 %5 %6 %7 %8 %9
+set z_ReturnCode=%ERRORLEVEL%
+if %z_ReturnCode% GTR 4 (erase %1.OBJ
+                         echo %0 ERROR: See errors on mz390 generated bal file and console
```
jyganci: The above call to MZ390.BAT already did the same echo.

bat/ASML.BAT
```diff
@@ -1,35 +1,48 @@
...
-    echo ASML     ERROR: See errors on lz390 generated lst file and console
-    erase %1.390
-    exit /b %RC%
-)
+call %~dps0lz390 %z_TraceMode% %1 %2 %3 %4 %5 %6 %7 %8 %9
+set z_ReturnCode=%ERRORLEVEL%
+if %z_ReturnCode% GTR 4 (erase %1.390
+                         echo %0 ERROR: See errors on lz390 generated lst file and console
+                         goto return
+                         )
```
jyganci: The above call to LZ390.BAT erased %1.390 and did the same echo.

bat/CBLC.BAT
```diff
@@ -1,43 +1,55 @@
...
 
 rem get the z390 directory
-set "zdir=%~dps0..\"
-for %%f in ("%zdir%") do set "zdir=%%~ff"
+set "z_HomeDir=%~dps0..\"
+for %%f in ("%z_HomeDir%") do set "z_HomeDir=%%~ff"
+
+call %~dps0ZC390 %z_TraceMode% %1 %2 %3 %4 %5 %6 %7 %8 %9
+set z_ReturnCode=%ERRORLEVEL%
+if %z_ReturnCode% GTR 4 (echo %0 ERROR: Cobol compile error
```
jyganci, line 44: Shouldn't "Cobol" be "COBOL"?

bat/CBLC.BAT
```diff
@@ -1,43 +1,55 @@
...
+                         goto return
+                         )
```
jyganci: Shouldn't "Cobol" be "COBOL"?

bat/ASMLG.BAT
abekornelis: Correct - I've changed accordingly

bat/ASM.BAT
abekornelis: I've modified to avoid duplicate messages

bat/ASML.BAT
abekornelis: messages are no longer duplicated

bat/CBLC.BAT
abekornelis: Wording has changed. Cobol/COBOL no longer in the message ...

dsh33782 (approved):
Abe, all
I'm not strongest bat developer, I do not currently have option to code or test bash equivalents. But I did look at all of these 43 bat file changes.  I trust your assurance that they all work as intended.

Issue #372, opened 2022-03-16
extend zCobol test suite with program to calculate Pi


abekornelis:
Program supplied by James Francis Cray

PR #373, opened 2022-03-17
Issue372 add new program to zCobol test suite

abekornelis:
There are two copies of the program. The YouzPi1 runs fine, but the YouzPi2 fails at runtime with a S0C7.
Too bad - but I do not have the time now to investigate and fix this issue, which is probably a compiler bug.
The problem can be addressed at a later stage if anyone decides it's worth the time and effort.

This request solve issue #372 

This pull request is dependent on the fix for Issue 363. This pull request to be reviewed after Issue 363 has been closed.


dsh33782 (approved):
18 changes look ok to me.

zcobol/test/YOUZPI2.CBL
```diff
@@ -0,0 +1,47 @@
...
+               VARYING COUNTER FROM 1 BY 1
+               UNTIL COUNTER GREATER THAN ITERATIONS.
+           MULTIPLY MYSUM BY 4.00000 GIVING MYSUM1.
+           DISPLAY ' YOUR YOU YOUZPI CALCULATED TO BE:'.
+           DISPLAY MYSUM1.
+           DISPLAY THANKS.
+           STOP RUN.
+
+       MYLOOP.
+           MULTIPLY COUNTER BY 2.0 GIVING TEMP1.
```
jyganci, line 41: Since this program is not part of a test, the following comments are documenting the S0C7 that occurs when attempting to run it.

This MULTIPLY is causing the S0C7. Reason follows.

Looking in YOU2PI2.PRN you will see that this MULTIPLY becomes
```
FC4390A8E000     MP    ZCVT_PWORK1(5),0(4,ZC_R14)
```
The value in ZCVT_PWORK1(5) is X'000010000C', which is 1, adjusted for the PICTURE used for TEMP1. The value in 0(4,ZC_R14) is X'0000020C', which is 2.0 (2 adjusted for one decimal place). The MP instruction will S0C7 if the first operand does not have 4 leading bytes of zeros -- the length of the second operand. As you can see, the first operand only has 2 leading bytes of zeros.

This is certainly a bug in the zCOBOL compiler. The MP instruction can get a specification exception (S0C6) if the multiplier (2nd operand) has length greater than 7 or the length of the multiplier exceeds the length of the multiplicand. It was easy to create test cases for each of these possibilities. For example, change the MULTIPLY instruction to "MULTIPLY COUNTER BY 2 GIVING TEMP1." 

YOUZPI2.CBL is the same as YOUZPI1.CBL except all numeric fields are zoned decimal -- an inefficient format when doing arithmetic calculations. For example, COUNTER is initialized to 1, which is X'F0F0F0F0F0'; FAC is initialized to 1 --- since it is signed, its value is X'C0'.

Lot of work for someone to do here.

jyganci (approved):
Looks good. As you noticed, the second COBOL program does not work on z390. Okay since it is not used in a test.

zcobol/test/YOUZPI2.CBL
abekornelis: I have converted your analysis into a new issue #376. Thanks for your detailed analysis!

PR #381, opened 2022-03-28
Change assembler directive from APARM to ACALLPRM to allow use of APARM macros

dsh33782:
This pull has been updated with correct date in log and suggested use of all caps for ACALLPRM.


jyganci (approved):
Looks good.

Issue #327, opened 2021-11-11
Fix LTORG pool to align odd length literals on even bound when referenced by LARL which uses half word offset


dsh33782:
Fix LITORG pool to align odd length literals on even bound when referenced by LARL which uses half word offset.  This issue has been raised as a result of discussion on the IBM Assembler email list group which reported recent APAR fix for HLASM:
>>>>>>>>>>>>>>>>>>
APAR PH34824 in March 2021 fixed the LARL loophole; if a literal with odd byte length is referenced by relative address, it is now automatically aligned to a halfword by inserting a padding byte before it if necessary. 
It is still stored within the last section of the literal pool (because at the time of definition, it is not necessarily known 
whether it will be referenced by relative address) so there may occasionally be a padding byte within the pool as a result. 
Jonathan Scott, HLASM IBM Hursley, UK
<<<<<<<<<<<<<<<<<<<<<<
[LARLLIT.MLC.TXT](https://github.com/z390development/z390/files/7516686/LARLLIT.MLC.TXT)
[LARLLIT.PRN.TXT](https://github.com/z390development/z390/files/7516687/LARLLIT.PRN.TXT)
<

PR #382, opened 2022-03-28
fix LTORG issue 327 to align odd literals to even boundary

dsh33782:
Issue #327 assember error due to LARL referencing odd address literal has been fixed by changing az390.java to force even addresss for all literals.  A new regression test tests\TESTLITS.MLC has been added to rt\bat\RUNASMTESTS.BAT.  The BUILD.BAT regression test has run successfully locally.

### DIFF
--- a/src/az390.java
+++ b/src/az390.java
@@ -413,7 +413,8 @@ public  class  az390 implements Runnable {
         * 2821-04-26 Issue #239 fix no error on undefined sym for RIL i2 operand
         * 2021-08-20 DSH issue #230 correct vector instruction erros reported by Dan Greiner
         * 2021-09-07 dsh #230 fix E7CC option, fix E7C0-E7C7 OR 8 with operand m4 or m6	
-        ^ 2022-01-16 dsh #343 move abort for exceeding maxline 		
+        * 2022-01-16 DSH #343 move abort for exceeding maxline 
+        * 2022-03-28 DSH #327 fix az390 to force odd literals to even address for access by relative halfword offset counts	
     *****************************************************
     * Global variables                        last rpi
     *****************************************************/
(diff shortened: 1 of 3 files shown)


src/az390.java
```diff
@@ -10589,6 +10590,8 @@ private void gen_ltorg(){
 private void gen_lit_size(int size){
 	/*
 	 * generate literal dc's of specified size
+           * force all length 1 lits to even addr for access by relative halfword offsets in LARL etc.  per issue #327
+           * see regression test rt\test\TESTLITS.MLC which fails on previous releases
```
jyganci, line 10594: Comment only. Shift the 2 lines left so the asterisks align with the asterisk on the preceding line.

src/az390.java
```diff
@@ -10598,6 +10601,9 @@ private void gen_lit_size(int size){
 				){			
 			lit_gen[cur_lit] = 1;
 			lit_esd[cur_lit] = esd_base[cur_esd]; // RPI 457
+                             if   (size == 1 && loc_ctr != (loc_ctr/2*2)){
+                                  loc_ctr = loc_ctr + 1;  // issue #327 force lits on even addr for ref by halfword offset counts
+                             }
```
jyganci, line 10606: Suggestion: Shift code left so that it aligns with existing code.

I make no claim about understandiing the az390 logic, but I do have a question about this modification. Why are you only checking for size 1? Why not any odd size?

Never mind. I see in az390 that size 1 is really a power of 2 and does not mean the length of the literal. Code looks fine.

jyganci (approved):
Looks good.

src/az390.java
jyganci: After updating my next comment below with regard to "size", maybe you should replace "force all length 1 lits ..." with "force all size 1 lits ..." on the above comment line.

dsh33782: I agree the size 1 is more descriptive as all the 8, 4, and 2 size literals have already been allocated and only the odd length literals are in size 1 group and the fix just forces those ending up on odd byte address to be on even byte address so LARL etc can access them.

Issue #374, opened 2022-03-18
Improve usability of bat files


abekornelis:
1) if name of file to be assembled/compiled/linked/executed is missing, issue prompt.
2) if prompted value is absent, issue explanatory text

PR #383, opened 2022-03-30
Issue374 Improve usability of bat files

abekornelis:
this pull request solves issue #374

### DIFF
--- a/bat/ASM.BAT
+++ b/bat/ASM.BAT
@@ -17,9 +17,16 @@ if /I %z_file%. == . goto help
 :chkfile
 rem convert Unix path notation to Windows before checking the file
 set z_file=%z_file:/=\%
-if not exist %z_file%.MLC (echo %0 ERROR: %z_file%.MLC was not found
-                           set z_ReturnCode=16
-                           goto return)
+if exist %z_file%.MLC goto file_ok
+rem if user specified valid extension, strip it off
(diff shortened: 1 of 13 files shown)


bat/ASM.BAT
```diff
@@ -15,9 +15,18 @@ if /I %1. == . (set /P z_file=Hit Enter for help or suppply name of mlc file to
         ) else (set z_file=%1)
 if /I %z_file%. == . goto help
 :chkfile
-if not exist %z_file%.MLC (echo %0 ERROR: %z_file%.MLC was not found
-                           set z_ReturnCode=16
-                           goto return)
+rem convert Unix path notation to Windows before checking the file
```
jyganci, line 18: Why? This is a Windows BAT file. Shouldn't it be a "JCL error" if a Windows user enters *nix file separators? Is there some reason for doing this?

abekornelis: I'm striving for platform independence. I'm aware it'll take a lot more than just this. We already had some issues with .opt files pointing where chaining another file had to be done differently depending on the platform being used. I think that sucks... It's easy to code - no big deal. Makes it just a bit easier again for the user.

dsh33782 (approved):
Looks good to me.  I appreciate help with URL's remembering if its / or \.

Issue #384, opened 2022-03-30
Add test to bash/runasmtests that was added to rt/bat/RUNASMTESTS.BAT for issue #327


jyganci:
Issue #327 "Fix LTORG pool to aling odd length literals on even boundary" add new test program TESTLITS.MLC and new run of it to rt/bat/RUNASMTESTS.BAT. Need to add corresponding run of it to bash/runasmtests.

jyganci:
Merged into base.

PR #385, opened 2022-03-30
#384 Add TESTLITS test to bash/runasmtests

jyganci:
Issue #327 "Fix LTORG ..." added a new test program TESTLITS.MLC and updated rt/bat/RUNASMTESTS.BAT to invoke it. This pull request is an update to bash/runasmtests to also invoke TESTLITS.


dsh33782 (approved):
Looks good.  Thanks.

abekornelis (approved):
I'm  missing the optable specification, but I guess it does not really make a difference for testing literals. DC/DS should be in all optables we support.
As far as I'm concerned, this change is okay

PR #395, opened 2022-04-30
Fix build error "Not a git repo."

adelosa:
The `Not a git repo` message has been causing the linux build job to fail.
After investigation, I tracked the cause to a Git security issue and subsequent patch that has been applied.

* https://github.blog/2022-04-12-git-security-vulnerability-announced/
* https://github.com/actions/checkout/issues/760

Once this is merged, it should address the build issues.


jyganci (dismissed):
Looks okay.

adelosa:
Added fix to RELVER.BAT to remove space from version in z390.properties.

jyganci (approved):
Changes look good. Also looked at bash/relver. It does not require change.

PR #397, opened 2022-05-03
Add sample program for issue 396 - Incorrect code generated for Branch without operands

abekornelis:
This program is the result of manually recreating the problem. Proper revision should be part of the review of the solution. For now, we just want the test program to go into the repository.


dsh33782 (approved):
Ok, example with missing B operand.

rt/mlc/IS00396.MLC
```diff
@@ -0,0 +1,26 @@
+TESTB    START 0
+         USING *,13
+         STM   14,12,12(13)
+         ST    13,8(13)
+         ST    15,4(15)
+         LR    13,15
+         B
+         B     
+         B     ,
```
jyganci, line 9: Why 3 instructions? The first 2 are the same???

Same comment for BNE, J, SVC, and MVC.

abekornelis: John, good question. The initial report was about a B with a single comma indicating absence of operands. I extended the test case to also omit the comma with or without trailing blanks after the B instruction - just to validate assembler behavior. Then I added the other instructions to fathom if they would suffer the same problem. Thus you see a lot of repetition of almost identical testcases.

PR #401, opened 2022-05-05
Sample program to reproduce/test issue 398 - Typo in error message 42 - invalid relatvie offset expression

abekornelis:
This program is the result of manually recreating the problem. Proper revision should be part of the review of the solution. For now, we just want the test program to go into the repository.


dsh33782 (approved):
OK.  It looks like a simple test of linkage to me.

jyganci (approved):

Issue #325, opened 2021-10-05
Assembler hangup on DC X'.....'....' 


dsh33782:
Following statement causes assembler to abort reported by Dan Greiner:

main    csect
        dc      x'12345678,98765432,55555555'78451256'
        end

PR #406, opened 2022-05-08
Sample 00403 - three test cases for issues reported by John Hennesy

abekornelis:
I created three new issues, based on reports by John Hennesy. Sample programs to reproduce the symptoms and for creating an RT script to validate the fix are in this pull request.


abekornelis:
Sample programs and expected results have been moved to their respective issues. No need to insert these programs upfront into the repository. They'll be added to the RT collection with the resolution of each issue.

dsh33782:
Abe, all

 

I think this is a much better approach to handling single error conditions, and I will do the same with the #325 assembler abort issue I’m working on today.  I will attach test program and expected result vs current abort error.  However, we may want to review each release and consider adding some sort of composite regression test or test frame case to verify multiple fixes in each release.

 

Don Higgins

 ***@***.***> ***@***.***

www.don-higgins.net

PR #427, opened 2022-06-24
Issue #423 correct assembly of rotate and insert ins

### LINKED ISSUE #423: Correct Rotate and insert, or, xor instruction before adding new z16 extentions of instruction
Bug reported by Dan Greiner:

While exploring the various new extended mnemonics that use the ROTATE THEN *** instructions, I discovered a bug in z390's implementation:

RNSBG, ROSBG, and RXSBG mnemonics have a "T" suffix that sets the test-results control in bit 0 of the I3 field. Similarly, RISBG, RISBGN, RISBHG, and RISBLG have a "Z" suffix that sets the zero-remaining-bits control in bit 0 of the I4 field. However, it appears that the z390 code to implement these suffixes adds a value of 128 to the specified field's value, rather than ORing a value of 128.

Since a programmer is not limited to specifying an I3 and I4 bit range between 0 and 63, its possible that a larger value could have been specified, causing the addition of 128 to wrap around, thus ending up with either the control bit being zero or a wrong value generated. Examples are shown below. This is especially troublesome in the case of the new SHIFT LEFT/RIGHT LOGICAL extended mnemonics that showed up in the z16 PoO.

000000 EC127FFFFF54                           (1/3)3         RNSBGT  1,2,255,255,255
000006 EC347FFFFF56                           (1/4)4         ROSBGT  3,4,-1,-1,-1
00000C EC56D7010057                           (1/5)5         RXSBGT  5,6,31-200,1,0 (should be 57, not D7)
000012 EC67087F0955                           (1/6)6         RISBGZ  6,7,8,255,9
000018 EC890A7F0B59                           (1/7)7         RISBGNZ 8,9,10,255,11
00001E ECCD0E7F0F5D                           (1/8)8         RISBHGZ 12,13,14,255,15
000024 ECEF007F0151                           (1/9)9         RISBLGZ 14,15,0,255,1

dsh33782:
Issue #423 correct assembly of rotate and insert ins option update with or(|) instead of add(+)

### DIFF
--- a/src/az390.java
+++ b/src/az390.java
@@ -417,7 +417,8 @@ public  class  az390 implements Runnable {
         * 2022-03-28 DSH #327 fix az390 to force odd literals to even address for access by relative halfword offset counts	
         * 2022-05-07 DSH #233 allow spaces within DC numberic values for BDEFHLPXZ such as DC F'123 456' same as F'123456'
 		* 2022-05-10 AFK #398 fix typo in error message
-    * 2022-05-12 DSH #325 issue error 54 invalid DC field terminator if not '..' for BDEFHLPXZ
+        * 2022-05-12 DSH #325 issue error 54 invalid DC field terminator if not '..' for BDEFHLPXZ
+		* 2022-06-10 DSH z16 #423 correct rotate instruction errors reported by Dan Greiner
 	*****************************************************
     * Global variables                        last rpi
     *****************************************************/
@@ -3553,7 +3554,7 @@ private void process_bal_op(){
     	break;
     case 52: // "RIE8" "RNSBG" oo12334455oo RPI 817
         // r1,r2,i3,i4[,i5] or extended op r1,r2 RPI 1164
-    	// i3 high bit for test RNSBGT
+    	// i3 high bit for test RNSBGT, RXSBGT
     	// i4 high bit for zero RISBGZ
     	bal_op_ok = true;
     	loc_ctr = (loc_ctr+1)/2*2;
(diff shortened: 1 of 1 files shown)


dsh33782:
This is first prerequisite to adding z16 instruction extentions which use these rotate and insert instructions.

src/az390.java
```diff
@@ -3553,7 +3554,7 @@ private void process_bal_op(){
     	break;
     case 52: // "RIE8" "RNSBG" oo12334455oo RPI 817
         // r1,r2,i3,i4[,i5] or extended op r1,r2 RPI 1164
-    	// i3 high bit for test RNSBGT
+    	// i3 high bit for test RNSBGT, RXSBGT
```
jyganci, line 3557: i3 is high bit for test RNSBGT, RXSBGT, ROSBGT

src/az390.java
```diff
@@ -3553,7 +3554,7 @@ private void process_bal_op(){
...
     	// i4 high bit for zero RISBGZ
```
jyganci, line 3558: i4 high bit for zero RISBGZ, RISBGZN

src/az390.java
```diff
@@ -3578,16 +3579,19 @@ private void process_bal_op(){
        	} else {
        		skip_comma();
        		get_hex_byte();   // I3
+			if (tz390.opt_traceall){ // DSHX debug RXSBGT
+				tz390.put_trace("RNSBG/RXSBG I3=" + exp_val + "obj_code=" + obj_code);
```
jyganci, line 3583: Change RNSBG/RXSBG to RNSBG/RXSBG/ROSBG.

jyganci (approved):
I agree that the code changes to use "or" instead of "plus" are the right thing to do. However, I am still concerned that the bits that are supposed to be zero are not set correctly. Will approve this anyway. This change is progress, but further work is needed.

PR #429, opened 2022-06-25
Convert zCobol documents to markdown

abekornelis:
First - preliminary - version.


adelosa:
When building the docs with mkdocs, received the following errors:

```
INFO     -  [macros] - ERROR # _Macro Syntax Error_
            _Line 92 in Markdown file:_ **Missing end of comment tag**
            ```python
            ## zCOBOL Compiler Commands {#Commands}
            ```
INFO     -  [macros] - ERROR # _Macro Syntax Error_
            _Line 261 in Markdown file:_ **Missing end of comment tag**
            ```python
            ## Trademarks {#Trademarks}
            ```
INFO     -  [macros] - ERROR # _Macro Syntax Error_
            _Line 195 in Markdown file:_ **Missing end of comment tag**
            ```python
            ## Trademarks {#Trademarks}
            ```
```

I'll see if can work out the issues.

There is a bit of work required there for this to be ready for inclusion in the docs.. but I think you already know this..
There should not be any references to version numbers or companies (besides the new z390 LLC)
I also see a lot of stuff that should probably be edited to make it the same tone as the rest of the docs.. 

I am marking this pull request as draft until its ready to review in full.

I think the first step is determining where you want the zCOBOL content in terms of the overall structure of the docs site.

I have detailed this in the following page:

https://z390development.github.io/z390/contributing/contribute_docs/

This page also has instructions for building the doco on your own machine so you can see what it looks like.

So please review the current docs to get a feel for the current docs and where this new content should live.

I don't really want a new section just for zCOBOL, but I open to discussion on this point. I think the content can live within getting started, user guide and reference sections.

I am happy to work with you with this one, but it would be good if I can just make changes as I find issues rather than spending more time documenting the issue here so you can fix.. We can do that 2 ways. You can grant me permission to your fork, or we move the branch to the main z390development repo and can both contributes.. Let me know which way you want to go.

If you want to see how linking works, review the other pages in doc folder that have linking that works.. For the indexes, you need to define the anchor or get it from the TOC.. but for now, don't stress about that.. Lets focus on the structure and then we can get the links right later.

adelosa (approved):
Happy to merge this as-is. Can evolve over time as any issues are discovered.

Issue #414, opened 2022-05-16
Add assembler support for all new z16 instructions


dsh33782:
This issue has been split into:
- #503 
- #504 
- #505 
- #506 
- #507 
- #508

dsh33782:
[z16_new_instructions_2022_05_20.pdf](https://github.com/z390development/z390/files/8742233/z16_new_instructions_2022_05_20.pdf)
[z16_new_instructions_2022_05_20.xlsx](https://github.com/z390development/z390/files/8742234/z16_new_instructions_2022_05_20.xlsx)

PR #430, opened 2022-07-09
add 30 z16 new instructions to assembler #414 plus missing NAN consta…

dsh33782:
I've added TESTNANS.TXT and TEST#414.TXT to test new NANS and 30 new z16 instructions.
I've sent copy to Dan Greiner who has created macros for the new instructions, but I've not heard back yet.
[TESTNANS.TXT](https://github.com/z390development/z390/files/9076438/TESTNANS.TXT)
[TEST#414.TXT](https://github.com/z390development/z390/files/9076446/TEST.414.TXT)

### DIFF
--- a/src/az390.java
+++ b/src/az390.java
@@ -419,7 +419,10 @@ public  class  az390 implements Runnable {
 		* 2022-05-10 AFK #398 fix typo in error message
         * 2022-05-12 DSH #325 issue error 54 invalid DC field terminator if not '..' for BDEFHLPXZ
 		* 2022-06-10 DSH z16 #423 correct rotate instruction errors reported by Dan Greiner
-	*****************************************************
+	    * 2022-06-27 DSH z16 #424 add BFP constants INF, NAN, QNAN, SNAN, DMIN
+	    *                             See POP 13 pages 9-5 and 19-4
+		* 2022-07-03 DSH #414 add 30 new z16 instructions also see
+		*****************************************************
     * Global variables                        last rpi
     *****************************************************/
 	tz390 tz390 = null;
@@ -1788,7 +1791,9 @@ private void gen_list_mnemonics() // Routine added for RPI 1209A
                     &&  !tz390.op_name[index].equals("RSCH")
                     &&  !tz390.op_name[index].equals("SCHM")
                     &&  !tz390.op_name[index].equals("SAL")
-                    &&  !tz390.op_name[index].equals("XSCH"))
+                    &&  !tz390.op_name[index].equals("XSCH")
+					&&  !tz390.op_name[index].equals("LBEAR") // DSH #414 z16
+					)
                        {entry=entry+" D2(B2)";
                         }
                     break;
@@ -2629,7 +2634,9 @@ private void process_bal_op(){
     	loc_len = 4;
     	get_hex_op(1,4);
     	get_hex_zero(2);
-    	if (bal_op.equals("PALB") // RPI 277
+		if (bal_op.equals("NNPA")){ // DSH #414 z16
+			get_hex_zero(2);
+		} else if (bal_op.equals("PALB") // RPI 277
     		|| bal_op.equals("PCKMO") // RPI 1125
 	        || bal_op.equals("PCC")){ // RPI 1125
     		get_hex_zero(2);
@@ -2815,8 +2822,12 @@ private void process_bal_op(){
     	loc_len = 6;
     	get_hex_op(1,2);
     	get_hex_bdddhh2();
-    	skip_comma();
-    	get_hex_byte();
+		if (!bal_abort && exp_next_char(',')){ // DSH #414 z16 add support for EB71 LPSWEY D1(B1) witn no second opr
+      	   skip_comma();
+    	   get_hex_byte();
+		} else {
+		   get_hex_zero(2);
+		}
     	obj_code = obj_code.substring(0,2) + obj_code.substring(8,10) + obj_code.substring(2,8); 
        	get_hex_op(3,2);
     	check_end_parms();
@@ -3023,13 +3034,17 @@ private void process_bal_op(){
     	loc_start = loc_ctr;
     	loc_len = 4;
     	get_hex_op(1,4);
-    	get_hex_reg();
+    	get_hex_reg(); // r1
     	skip_comma();
-    	get_hex_reg();
+    	get_hex_reg(); // r3
 		skip_comma();
-    	get_hex_reg();
-    	skip_comma();
-    	get_hex_reg();
+    	get_hex_reg(); // r2
+		if (!bal_abort && exp_next_char(',')){ // DSH #414 z16 make M4 optional for RDP
+			skip_comma();
+			get_hex_reg();  // M4
+		} else {
+    	    get_hex_zero(1);
+		}
        	obj_code = obj_code.substring(0,4)  // oooo 
         + obj_code.substring(5,6)  // r3
 		+ obj_code.substring(7,8)  // m4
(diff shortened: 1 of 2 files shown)


src/az390.java
```diff
@@ -10923,6 +10954,196 @@ private void fp_get_hex(){
...
+				break;
+			case 8: // tz390.fp_lh_type s1,e7,m112 with split hex	
+				dc_hex = "00000000000000007200000000000001";
+				break;
+			case 9: // tz390.fp_lq_type quad word RPI 1108
+				dc_hex = "00000000000000000000000000000000";
+				break;	
+			}
+			if (fp_sign == '-'){
+				dc_hex = "8" + dc_hex.substring(1);
```
abekornelis, line 11195: In this block, please replace all the literal magic numbers by named constants. And please use names that match the naming conventions used in the PoP

src/tz390.java
```diff
@@ -1710,7 +1711,7 @@ public  class  tz390 {
      String[]   op_table_370_notsupported = // Table added for RPI 1209A
         {"RIO      S    9C02 D2(B2)",
          "CLRCH    S    9F01 D2(B2)",
-         "CONCS    S    B200 D2(B2)",
+         // DSH #414 new z16 LBEAR "CONCS    S    B200 D2(B2)",
```
abekornelis: You've accidentally(?) commented out the entry for CONCS. 
LBEAR does not belong in this table: it never was a 370 instruction
It is correctly defined in op_table_z16.

src/tz390.java
```diff
@@ -3841,6 +3875,25 @@ public void create_opcodes()  // Routine added for RPI 1209
          process_opcodes(op_table_ZS3);
          process_opcodes(op_table_ZS4);
          process_opcodes(op_table_Z15);  // rpi 2202
+         process_opcodes(op_table_z16); // DSH #414
```
abekornelis: z16 opcodes should not be included when machine level is z15

src/tz390.java
```diff
@@ -4892,6 +4952,7 @@ else if (opt_machine.contentEquals("Z15"))
                &&  opt_optable.equals("ZS3") != true
                &&  opt_optable.equals("ZS4") != true
                &&  opt_optable.contentEquals("Z15") != true  // rpi 2202
+               &&  opt_optable.contentEquals("Z15") != true  // DSH #414
```
abekornelis: z15 should be z16 ??

src/az390.java
dsh33782: These constants don't have any names that I can find.

src/tz390.java
dsh33782: CONCS is being removed because the same opcode is now new z16 LBEAR opcode

src/tz390.java
dsh33782: fixed.

dsh33782:
I've added second commit for same 2 files to fix a few z16 instruction formats.

src/az390.java
```diff
@@ -2815,8 +2823,12 @@ private void process_bal_op(){
...
     	get_hex_op(1,2);
     	get_hex_bdddhh2();
-    	skip_comma();
-    	get_hex_byte();
+		if (!bal_abort && exp_next_char(',')){ // DSH #414 z16 add support for EB71 LPSWEY D1(B1) witn no second opr
+      	   skip_comma();
+    	   get_hex_byte();
+		} else {
+		   get_hex_zero(2);
+		}
```
jyganci, line 2831: First off, I make no claim as to understanding this code. 

Old code did skip_comma(), get_hex_byte(), then replaced obj_code with 3 substrings.

New code added the "if else" before the replacment of obj_code. The "else" is taken when bal_abort is true or the next claracter in the expression is not a comma, concatenating 2 bytes of "0" to obj_code. Now the replacement of obj_code is done.

Just being sure this is what you intended.

Same comment applies below for your new lines 3042-3048.

dsh33782:
John

 

Thanks for reviewing these changes.

 

Yes the LPSWEY has no second operand and just plugs in zeros.

 

Don Higgins

 ***@***.***> ***@***.***

www.don-higgins.net

src/az390.java
```diff
@@ -10923,6 +11005,196 @@ private void fp_get_hex(){
...
+			    break;
+			case 1: // tz390.fp_dd_type s1,cf5,bxcf8,ccf50
+				dc_hex = "EEEEEEEEEEEEEEEE"; 
+				break;
+			case 2: // tz390.fp_dh_type s1,e7,m56 with hex exp
+				dc_hex = "EEEEEEEEEEEEEEEE";
+				break;
+			case 3: // tz390.fp_eb_type s1,e7,m24 with assumed 1
+	            dc_hex = "7F000000";
+	            break;
```
jyganci, line 11022: According to tz390, case 3 is for BFP short format. According to PoP 13, p19-2, the exponent is 8 bits (bits 1-8), so the value "7F000000" should be "7F800000".  Also, don't know what you mean by "with assumed 1" in the comment on line 11020.

src/tz390.java
```diff
@@ -3841,6 +3875,25 @@ public void create_opcodes()  // Routine added for RPI 1209
          process_opcodes(op_table_ZS3);
          process_opcodes(op_table_ZS4);
          process_opcodes(op_table_Z15);  // rpi 2202
+         // no z16 for z15 process_opcodes(op_table_z16); // DSH2 #414
+         process_opcodes(op_table_DOS_directives);
+         process_opcodes(op_table_370_directives);
+         }
+         if ((opt_optable.equals("Z16"))
+            |(opt_optable.equals("ZSA")))  // DSH #414
```
jyganci, line 3883: |(opt_optable.equals("ZSA")))

should be

||(opt_optable.equals("ZSA")))

src/az390.java
abekornelis: we can assign them. I'm much in favor of following IBM terminology. But if not provided, we're free to design our own in z390. I suggest [pos|neg]_[minval|maxval|infinity]_[db|dh|ed|eh|lb|ld|lh|lq]
Please modify or extend as needed...

src/tz390.java
```diff
@@ -1710,7 +1711,7 @@ public  class  tz390 {
      String[]   op_table_370_notsupported = // Table added for RPI 1209A
         {"RIO      S    9C02 D2(B2)",
          "CLRCH    S    9F01 D2(B2)",
-         "CONCS    S    B200 D2(B2)",
+         // DSH #414 new z16 LBEAR replaces old nonsupported "CONCS    S    B200 D2(B2)",
```
abekornelis, line 1714: LBEAR is defined in the table for z16 instructions. Please leave CONCS as it was. We want to keep track f unsupported instructions and their position in the sequence of PoP manuals. That should obviate the need to go find all the details in case we need to implement the CONCS instruction...

src/az390.java
```diff
@@ -10923,6 +11005,196 @@ private void fp_get_hex(){
...
 			return;
+
+		} else if (fp_text.toUpperCase().equals("(INF)")){ // #423
+			switch (tz390.fp_type){  // gen (min) hex for tz390.fp_type
+			case 0: // tz390.fp_db_type s1,e11,m52 with assumed 1
+				dc_hex = "7FF0000000000000";
+			    break;
+			case 1: // tz390.fp_dd_type s1,cf5,bxcf8,ccf50
+				dc_hex = "EEEEEEEEEEEEEEEE"; 
+				break;
```
jyganci, line 11016: 1. INF, NAN, QNAN, and SNAN only apply to binary and decimal floating point. See the HLASM Language Reference manual; p156 of the -9 version. It also shows that DMIN is another special value. MAX, MIN and DMIN apply to all 3 floating point types. The az390 code does NOT implement "(DMIN)". Should it?
2. Why all the "EEE...EEE" values? Why not the correct values? John Ehrman's assembler text has tables that show the exact values. For example, the value for INF for DD is "7800000000000000". See Table 325 in Section 33.5 of the book for all the DFP values. See Table 276 of Section 34.2 for BFP values. See Table 238 in Section 33.4.2 for the MAX, MIN and DMIN values for HFP. Note there are no HFP INF, NAN, QNAN or SNAN values.
3. Unrelated, but some of the MIN values for HFP are incorrect and the MIN values for DFP appear to be DMIN values. Please doublecheck all values against the tables mentioned above.
4. Not sure why you added INF,NAN, QNAN and SNAN for HFP. Shouldn't these generate assembly errors?
5. You should replace eachl "EEE...EEE" value with the corresponding correct value.

src/az390.java
```diff
@@ -419,7 +419,11 @@ public  class  az390 implements Runnable {
 		* 2022-05-10 AFK #398 fix typo in error message
         * 2022-05-12 DSH #325 issue error 54 invalid DC field terminator if not '..' for BDEFHLPXZ
 		* 2022-06-10 DSH z16 #423 correct rotate instruction errors reported by Dan Greiner
-	*****************************************************
+	    * 2022-06-27 DSH z16 #424 add BFP constants INF, NAN, QNAN, SNAN, DMIN
+	    *                             See POP 13 pages 9-5 and 19-4
+		* 2022-07-03 DSH #414 add 30 new z16 instructions also see
```
jyganci, line 424: End of line is " also see"/ Also see what???

dsh33782:
John

 

Before I launch into significant changes to the BFP and DFP z390 code to correctly label each type hex field, I’d like to verify that I have the latest book reference by John Ehrman: http://zseries.marist.edu/enterprisesystemseducation/assemblerlanguageresources/Assembler.V2.alntext%20V2.00.pdf

This is the only book I’m aware of that defines the hex values for limits on floating point constants of all types.

 

Don Higgins

 ***@***.***> ***@***.***

www.don-higgins.net

jyganci:
Don,

The book in the link seems to match my downloaded copy.

Regards,

John Ganci

PR #440, opened 2022-08-22
#000 Normalize whitespace

FredvanderWindt:


adelosa:
The decision to clean up whitespace is a large change to the codebase that will require the approval from the core contributors. It impacts many modules and will cause a mass merge issue for anyone who has work in progress.
@dsh33782 @abekornelis @jyganci -- Can I get your thoughts on this request. I am not against it but it would be nice to have a discussion about code standards and what linting tools we want to use to protect the source base.

@FredvanderWindt - I can see that the 2 fixes you have submitted are built on this request so they cannot be reviewed until we get though this change first.

dsh33782:
I need to learn more about what the impact of this change will be going forward before I can decide.

adelosa:
Hi @FredvanderWindt - The general view from the team is that such a change is not ideal.. We would prefer such changes are discussed and agreed before progressing. At this stage, the view is this should not go ahead. 
There is still interest in the fixes that came after this so I have created a new pull request with just the changes without the whitespace update in PR #444. Please check that all looks good.. Closing this item off without merge.

FredvanderWindt:
Hi,

Whitespace changes should not cause any merge conflicts in git.

But you are right that a ‘way of working’ with regards to whitespace should be chosen by the core team. Otherwise spurious white spaces will sneak into the code.

Generally the following approach is chosen:

* No trailing whitespace (all trailing whitespace is trimmed).
* Standardize leading whitespace (choose either spaces or tabs).

That last choice often depends on the programming language in the file: for COBOL you usually choose leading spaces, for languages like java or C you often see tabs.

But you should choose an approach and stick to that. The current sources are a bit messy in that regard.

Groetz,

Fred!

Issue #439, opened 2022-08-17
INDEX  issue


FredvanderWindt:
The INDEX function is not properly parsed when it is used in an expression.

Example:

```
&D       SETA     (((INDEX('!NO !YES','!&DCR')+3)/4)-1)
         AIF      (&D EQ 0 OR &D EQ 1).DCROK
         MNOTE    8,'FQ35 Characters decorate option "&DCR" invalid.'
         MEXIT
.DCROK   ANOP

```

Returns an error:

` *MZ390E error  38        (14/45)45 expression parsing error - prev op =IX next op =+
` 

**Proposed fix**

This can be fixed in exp_action array in mz390.java.

**Original**

 	int[] exp_action = {  
			1, 3, 3, 1, 0, 1, 1, 8, 1, 3, 1, 1, 1, 1, 3, // 1 +-  prev add/sub
			2, 2, 3, 2, 0, 2, 2, 8, 2, 3, 2, 2, 2, 2, 3, // 2 * / prev mpy/div  RPI 214
			3, 3, 3, 4, 3, 0, 3, 3, 3, 3, 3, 3, 3, 3, 3, // 3 (   prev open (...) RPI 274, RPI 647
			3, 3, 3,11, 3, 3, 3, 8,11, 3, 3, 3, 3, 3, 3, // 4 )   prev var subscript RPI 559
			0, 0, 3, 5, 5, 5, 5, 8, 0, 3, 0, 0, 0, 0, 3, // 5 .   prev concat
			3, 3, 3, 6, 3, 6, 3, 8, 3, 3, 3, 3, 3, 3, 3, // 6 ~   prev terminator
			3, 3, 3, 7, 3, 7, 7, 8, 0, 3, 7, 7, 7, 7, 3, // 7 LL  logical compares //RPI144 (was 1,2 now 3,3)
			3, 3, 3, 0, 0, 0, 0, 8, 0, 3, 0, 0, 0, 0, 3, // 8 '   string '....'
			3, 3, 3,10, 0, 0, 0, 0, 9, 3, 0, 0, 0, 0, 3, // 9 ,   substring '...'(e1,e2)
			12,12, 3,12,12,12,12, 3,12,03,12,12,12,12, 3, //10 ?'  prefix operator  //RPI145, RPI196, RPI 353 rpi 938 O3 FOR DOUBLE ?' etc
			3, 3, 3,13, 3, 0, 3, 3, 0, 3,13,13,13,13, 3, //11 NOT logical
[...]

PR #441, opened 2022-08-22
#439 index issue

FredvanderWindt:
Fix for issue #439: expressions with INDEX function and any other operation (add, subtract etc) fail with a parse error.


adelosa:
Rejecting this in favour of #444 which just contains the change without the whitespace update.

Issue #438, opened 2022-08-17
X2C issue


FredvanderWindt:
X2C function only works correctly for one byte hexadecimal string length input. X2C and X2B functions fail on multi-byte input data.

Example:

`&SRC     SETC  X2C('30313233343536373839616263646566414243444546')`

Returns an error, during macro expansion process:

```
14:44:20 FQM8921C  MZ390 START USING z390 V1.7.07 ON J2SE 1.8.0_333 05/12/22
MZ390E abort 84 file=1 line=43 internal system exception - java.lang.NumberFormatException: For input string: "30313233343536373839616263646566414243444546"
AZ390E error 165       (1/43)42   * MZ390E abort 84 file=1 line=43 internal system exception - java.lang.NumberFormatException: For input string: "30313233343536373839616263646566414243444546"
14:44:24 FQM8921C  MZ390 ENDED   RC=16 SEC= 3 MEM(MB)= 74 IO=181
```

**Proposed fix**
This can be fixed in function exec_pc_x2c in mz390.java.

**Original**

```
 private void exec_pc_x2c(){
    	/*
    	 * convert hex string to char string
    	 */
    	check_setc_quotes(1); // RPI 1139
    	setc_value1 = get_setc_stack_value();
    	seta_value = Integer.valueOf(setc_value1,16);
		setc_value = ""
	       + (char)tz390.ebcdic_to_ascii[seta_value >>> 24]
	       + (char)tz390.ebcdic_to_ascii[seta_value >>> 16 & 0xff]         
	       + (char)tz390.ebcdic_to_ascii[seta_value >>> 8  & 0xff]
	       + (char)tz390.ebcdic_to_ascii[seta_value        & 0xff]		       ;
	put_setc_stack_var();
    }
```

**Fixed**

```
    private void exec_pc_x2c(){
    	/*
    	 * convert hex string to char string
    	 */
    	check_setc_quotes(1); // RPI 1139
    	setc_value1 = get_setc_stack_value();
//ING    	seta_value = Integer.valueOf(setc_value1,16);
//ING	setc_value = ""
//ING	       + (char)tz390.ebcdic_to_ascii[seta_value >>> 24]
//ING	       + (char)tz390.ebcdic_to_ascii[seta_value >>> 16 & 0xff]         
//ING	       + (char)tz390.ebcdic_to_ascii[seta_value >>> 8  & 0xff]
//ING	       + (char)tz390.ebcdic_to_ascii[seta_value        & 0xff]	 
[...]

PR #442, opened 2022-08-22
#438 X2C issue

FredvanderWindt:
Fix for #438 X2C issue: SETC does not handle hex literals longer than two characters (one result byte) correctly.


adelosa:
Rejecting this pull request and have moved the X2C fix to PR #444.

Issue #443, opened 2022-09-16
Changes to bash scripts that incorporate corresponding changes to bat files


jyganci:
Issues #359, #363, and #374 modified some Windows bat files to better handle error conditions. The corresponding bash scripts should also be modified.

adelosa:
@jyganci Is this linked to a pull request you are working on?

jyganci:
Long time ago. I believe this is already resolved. See
https://github.com/z390development/z390/pull/445

PR #445, opened 2022-09-25
incorporate changes to corresponding bat files #443

jyganci:


bash/mac
```diff
@@ -50,11 +72,15 @@ if [ $debug -eq 1 ]; then
...
 rc=$?
 if [ $debug -eq 1 ]; then echo "mac: mz390 rc=$rc"; fi
-if [ "$rc" -ne "0" ]; then
-    echo "mac: mz390 error; see errors on mz390 generated $1.BAL file and console"
+if [ $rc -ne 0 ]; then
+    if [ $rc -eq 4 ]; then
+        echo "mac WARNING: mz390 rc=$rc; see warnings in mz390 generated $zFile.BAL/ERR file(s) and on console"
+    else
+        echo "mac ERROR: mz390 rc=$rc; see errors in mz390 generated $zFile.BAL/ERR file(s) and on console"
+    fi
```
adelosa, line 83: TIL: rc=0 means OK, rc=4 means warnings, rc=<any other value> means error. Personally I would prefer a range here like 0=OK,1-4=Warn, 5+=Error even though I suspect that values other than 0, 4 and some other value are the expected values.. Just a comment for thought..

bash/asm
```diff
@@ -5,25 +5,50 @@
...
+            shift
+            trace=1
+            ;;
+        "troff")
+            shift
+            trace=0
+            ;;
+    esac
+    # restore nocasematch setting
+    $SHELLNOCASEMATCH
```
adelosa, line 26: This logic is repeated in many of the modules -- just a thought.. is it worth having a shared library of bash functions that are includes as part of the individual commands so that we don't get this repetition?? So new bash script called z390bash.sh -- and each script sources this file -- source z390bash.sh. This then makes any functions defined available to the bash script.. Again.. happy to approve, just a thought for next iteration.

bash/ez390
```diff
@@ -25,6 +25,9 @@ fi
 java -classpath ${jardir}/z390.jar -Xrs $J2SEOPTIONS ez390 $1 $2 $3 $4 $5 $6 $7 $8 $9
 
 rc=$?
+# executed program: rc = any value (R15 at exit; set in sz390)
+# ez390: rc = 1 for Java issues; 16 for GUAM GUI window closed due to timeout
+#        rc = 16 for execution errors; 4,8 when using TEST (set in sz390)
```
adelosa, line 30: Comment on RC here but not acted on.. were you going to add some output based on the result??

bash/lz390
```diff
@@ -25,6 +25,7 @@ fi
 java -classpath ${jardir}/z390.jar -Xrs $J2SEOPTIONS lz390 $1 $2 $3 $4 $5 $6 $7 $8 $9
 
 rc=$?
+# rc = 1 for Java issues; all other non-zero values for linker issues
```
adelosa, line 28: Were you going to print something based on the rc.. or just add a comment in case someone looks through the script?

bash/mz390
```diff
@@ -25,6 +25,7 @@ fi
 java -classpath ${jardir}/z390.jar -Xrs -Xms150000K -Xmx150000K $J2SEOPTIONS mz390 $1 $2 $3 $4 $5 $6 $7 $8 $9
 
 rc=$?
+# rc = 1 for Java issues; all other non-zero values for assembly issues
```
adelosa, line 28: message based on rc?

adelosa (approved):
@jyganci - No real issues with these changes. I think there is some scope to consolidate some of the repetition through the modules by having a shared function but happy to progress as is and consider this in next iteration.

bash/asm
jyganci: Agreed, a lot of repetition. Didn't know about the sourcing technique. Maybe one day. However, I like seeing everything in one place.

While you didn't mention it, there is also a lot of duplication when writing error messages. For example, asm, asml, and asmlg duplicate error messages for the assembly step; asml and asmlg duplicate error messages for the link step. This was intentional. See my new comments below for ez390.

bash/ez390
jyganci: No. This is just documentation here. Partially followed what Abe did in the bat files. See xZ390.BAT (x=A,E,L,M). Abe chose to issue messages in xZ390.BAT and not in the bat files that invoke it. I chose to have the invokers of xz390 (x=E,L,M) issue messages. My reasoning was that the vast majority of users would use asm, asml, asmlg, cblc, cblcl, cblclg, exec, link and mac and would not directly invoke az390, ez390, lz390, or mz390; therefore, they should be the ones producing  any output messages.

bash/lz390
jyganci: No. This is just documentation. See my comment for ez390.

bash/mac
jyganci: IBM has used return codes that are multiples of 4 since the System 360 days. The z390 code also does.

I think the main reason was that a return code received in a calling program could then be used to "branch to a branch table entry" via "B    BRTAB(R15)". Register 15 contains the return code (a multiple of 4), and the code at BRTAB is a sequence of branches: like "B    RC0CODE", "B     RC4CODE", etc. The appropriate code is at each of the labels. A typical use might be
```
         ...
         L     R15,=A(SUB)     Address of subroutine
         BALR  R14,R15         Call subroutine
         B     BRTAB(R15)      Continue according to RC
BRTAB    DS    0H
         B     RC0CODE         Return code 0 processing
         B     RC4CODE         Return code 4 processing
         ...
```
This technique saves the caller from having to test each return code and continue accordingly - just 2 instructions are needed.

bash/mz390
jyganci: No. Just documentation. See my comment for ez390.

Issue #454, opened 2022-10-14
Questions concerning non-printable characters in PRN files


jyganci:
Mainframe HLASM SYSPRINT output shows non-printable characters as periods. For example, TSO edit for TA01B shows
```
-CAUTION- Data contains invalid (non-display) characters. Use command
          ===> FIND P'.' to position cursor to these                 
TA01B    CSECT 
ECFFC4444CECCE4444444
310120000325330000000
---------------------
         DC    C'A B'
444444444CC4444C7C0C7
0000000004300003D112D
---------------------
         END 
444444444CDC444444444
000000000554000000000
---------------------
```
The character between the A and the B, shown as space in TSO edit, is X'01'. When assembled, the return code is 0 and the SYSPRINT output shows the non-printable X'01' as a period.

```
  Loc  Object Code    Addr1 Addr2  Stmt   Source Statement 
000000                00000 00003     1 TA01B    CSECT       
000000 C101C2                         2          DC    C'A.B'
                                      3          END         
```
Using z390 to assemble TA01B yields
```
Assembler Listing
000000                                        (1/1)1 TA01B    CSECT 
000000 C101C2                                 (1/2)2          DC    C'AB'
000003                                        (1/3)3          END 
```
The error message
```
MZ390E error 138         (1/2)2 invalid ascii source line 2 in /<path>/TA01B.MLC
```
appears and the assembly ends with RC 12. Note the non-printable character between the A and the B.

If you look at Pull Request #444, you will see a sample program, TX2C1.MLC, that creates DC statements. using X2C. Modifying it to test X2C('C101C2' produces the same DC statement as above but the assembly ends with RC 0.
```
Assembler Listing
000000                                        (1/1)1 TX2C1    CSECT 
000000                                       (1/17)2 *        Test X2C with X'C101C2'
000000 C101C2                                (1/19)3          DC    C'AB'
000003                                       (1/20)4          END 
```
Questions:
1. Should z390 allow non-ASCII in a source  file?
2. Should z390 replace non-printable characters in PRN files with periods, as is done on the mainframe?

abekornelis:
Whether or not valid ASCII input text should be enforced is determined by the CHKSRC parameter. From the current doco:

> Check input source files as follows:
> 0 - no checking
> 1 - check MLC or BAL input source files for any  non-ASCII characters and issue error
> 2 - check MLC, MAC, CPY, and BAL input source files for any non-ASCII characters and issue error.
[...]

Issue #451, opened 2022-10-02
z390 ignores the CODEPAGE option for input files


jyganci:
An issue with the CODEPAGE option being ignored was posted to the z390 group by Martin Ward. 

Overview: 
1. An assembler source file has the character 0xA6 for a broken vertical bar, which is a valid character for the ISO-8859-1 encoding.
2.  When assembling the assembler source file, this should come out as X'6A' in the PRN file (EBCDIC encoding IBM1047 for a vertical bar) and the broken vertical bar should be displayed.
3. Using z390 built/run with Java 8, the assembler listing has X'8D' in the PRN file, not X'6A'. Also the character is not a displayable character.
4. Using z390 built/run with Java 19 (Martin's case), the PRN file has the three character sequence 0xEF, 0xBF, 0xBD which is the UTF-8 "replacement character" used when a character is not recognised.
This is what Java gives when a character 0xA6 is encountered when reading a UTF-8 character stream.
5. A quick fix to mz390.java (see Martin's second email and the response to it below) does partially solve the problem of properly handling of the 0xA6 character. When using z390 built/run with Java 8, X'6A' and the broken vertical bar are displayed in the PRN file. However, this is not the case for using z390 built/run with Java 19. A similar fix is required (see Martin's third email below) when writing the PRN file.

Summary:
1. The z390 software ignores the CODEPAGE option for input files.
2. The problem can be easily fixed for versions of Java less than 19.
3. Java 19+ will require changes to z390 to use CODEPAGE for output files.
4. All z390 Java source must be examined to see if there is any impact.

For documentation purposes, three of Martin's emails (and a response to one of them) follow.

********************************************************************************
Martin's first email
********************************************************************************
Summary: z390 seems to ignore the COPDEPAGE option for input files.

I recently updated to Java 19 from Oracle:

https://download.oracle.com/java/19/latest/jdk-19_linux-x64_bin.tar.gz

One of my assembler source files has the character 0xA6 for a broken
vertical bar, which is a valid character for the ISO-8859-1 encoding.
This should come out as X'6A' in the listing (EBCDIC encoding IBM1047
for a vertical bar). Instead, with the new version of Java I get
the three character sequence 0xEF, 0xBF, 0xBD which is the UTF-8
"replacement character" used when a character is not recognised.
This is what Java gives when a character 0xA6 is encountered
when reading a UTF-8 character stream.

I don't think z390 is honouring the CHARSET option for the input.

I have a file TEST.MLC which just declares a single vertical
[...]

Issue #447, opened 2022-09-27
Synchronize the bash, bat directories; the rt/bash, rt/bat directories; and build.sh, BUILD.BAT


jyganci:
Over time the contents of bash and bat directories have gotten out of sync due to changes/additions/subtractions to one of them not being propagated to the other. For example, a previous change moved several files from the bat directory to the rt/bat directory but did not move the corresponding bash files. The following changes are made by this issue:
1. move bash files corresponding to already moved bat files from the bash directory to the rt/bash directory
2. add new rt/bash files corresponding to existing rt/bat files
3. synchronize the build.sh script and BUILD.BAT file to invoke the same set of tests.

jyganci:
Merged by PR #450

PR #456, opened 2022-10-30
Use CODEPAGE option and handle non-printable chars

jyganci:
Martin Ward recently reported that z390 was not using the CODEPAGE values for input (and output). Issue #451 addresses this issue. In addition, it addresses issue #454 concerning non-printable characters being written to PRN, BAL and PCH files. One long outstanding issue is the "scraggly" lines in SNAP output when some of the characters in the "printable character" area of the output contain non-printable characters. The "fix" in the past was to use NOCODEPAGE. This has also been fixed here. Now one can dump storage using SNAP, look at storage in a dump, and display storage in test mode without having to specify NOCODEPAGE.

A new z390 option, WRITENONPRINTABLE, is being added that controls whether non-printable characters are written to PRN, BAL or PCH files. The default value, NOWRITENONPRINTABLE, replaces all non-printable characters with '.' while WRITENONPRINTABLE will write all characters to PRN, BAL and PCH files.

I made some decisions that require consensus. Some reasons/observations follow.
1. While not necessary for all files, there will be times when you will have to specifically set a code page value when opening a source file or an output file since each may have characters that are not displayable with whatever your editor uses for a code page value. For example, the default z390 CODEPAGE(ascii+ebcdic) option is CODEPAGE(ISO-8859-1+IBM1047). If you type in an "unusual" character in your source -- say U+00A6, which is "broken vertical bar" in ISO-8859-1 -- it may not display properly if you haven't specifically selected ISO08859-1 when opening the file.
2. Input to the assembler may come directly from a source file, be brought in by COPY, or be read in by AREAD. Output may be normal output written to PRN or BAL files, or via PUNCH to PCH files.
3. The replacement of non-printable characters with '.' is a compromise, modeled after what one sees when browising data on TSO. The "real" data is there, just shown as '.'. Turning "hex on" allows one to see what is actually there. We do not have that option in z390. The written value is either the actual character or '.'. Will BAL or PCH files need the actual data written to them, or is it okay to replace the non-printable characters with '.'?
4. This initial use of CODEPAGE only handles ISO-8859-1, UTF-8 or US-ASCII for ASCII character values and IBM1047 or IBM037 for EBCDIC character values. More may need to be added in the future.

I am looking for guidance here. Any help will be appreciated.


adelosa:
Hi John, some amazing work here. It will take me some time to review and figure out what is happening here before I could add much value to this.. I think the dot makes total sense where the real value is also provided as a hex value like in a dump..
```
F0F1 0000  2020 2020  0000 0000  2020 2020   [01..    ....    ] 
```
So with the above you have the hex and the printed version.. and 00 is not printable, so you get a dot but know that it was actually 00. 

In this case, having it return with the actual data does not make sense.

For cases where the hex display is not provided, then this option makes sense.. just trying to think when this happens because if an unprintable char is added in source, then presumably the full value would show in the instruction display which is in Hex.

Second point -- I really need to get my test framework merged so you don't need to create 2 scripts (bash/bat) and can add code that... 
```
Verify tests/T1P#451.BAL and tests/T1#451 look okay"
Verify console SNAP output does not contain scraggly lines"
```

adelosa (commented):
@jyganci -- A couple of things from me. 
* Can we add the new runcodepagetests tests to the build jobs. That way, we are adding regression to the build.
* I am not loving the Windows/Mac+Unix split in the test program. Can the filenames be passed in as params (maybe envvar) rather that having a Win/Unix flag.. that way the code is the same and the app calling the test is responsible for providing the file name in an OS specific way.. Specific filenames in the test cases is painful when coding in Groovy (see my pull request)
* I would generally not include binary data in the repo.. I would generally create the binary data as part of the test.. For example, if you wanted the char represented by a specific hex value, I would have the char hex value and create the binary value in the test cases so all code was readable.. 
* Doc update looks good..

I am still learning asm and macros so will take me quite some time to understand what the test program is checking. Based on how meticulous you are, I have no doubt they do the job well.

Let me know what you think on the above comments - they are my views, so happy to hear yours..

jyganci:
Responses to your bullet items.

1. Can we add runcodepagetests to the build jobs. Answer: yes. Reasons why
I did not do this: (1) not sure the pull request would be approved -- it is
large and a major change; (2) did not want to impact anyone else updating
the build jobs; (3) PR for issue #447 makes a lot of changes to the build
jobs but PR for issue #451 only adds one line; whichever PR gets merged
first affects the other one;  (4) being lazy about having to do a local
"git rebase" and "git force push" if someone else does update the build
scripts. I am assuming that PR for #447 will be merged before PR for #451.
After PR for #447 is merged I will add a commit to put it in the build jobs.

2. Windows/Mac+Linux split. I assume you are referencing T1#451.MLC lines
37-53. Note line 34, "AGO   .SKPCHK", skips over these lines. Note the
comment on the "AGO   .SKPCHK" line. Originally I used those lines but also
did not like having them so I skipped over them and tested
bash/runcodepagetests on Linux and bat\RUNCODEPAGETESTS.BAT on Windows.
Both ran successfully, even with the "hard-coded" Linux/Mac file names.
This verified that z390 took care of fixing the file separators when
running on Windows. I should have deleted these lines. I will add a commit
to do so.

3. Not including binary data in the repo. I assume you are referring to all
the ISO-8859-1 characters in the T1* files. I do understand what you mean
about creating the files, but this would have to be done outside of z390
since the files are used to verify z390 functionality. I will have to think
about this one.

Regards,

John Ganci

adelosa:
Hi @jyganci -- What is the status of this PR? Does it need to be reviewed in light of the new test framework which has now been merged or any other changes that have been made..

jyganci:
The status of the PR is waiting for approval. It's been quite some time since it was submitted, so I am not sure how/if the new test framework affects it. Will at least need to do another rebase to sync it with the latest z390 repo.

adelosa:
Hey @jyganci, I have not forgot about this and have spent a few hours looking at this today trying to think about the best approach to progress.

I like simplicity. Having a separate Java program compile a program to create test files seems complicated to me.. 
Now I can already hear you cursing me.. you will say "Anthony, you said I should have the files created by the test". 
This is true, and I do apologize as I think we had different understandings of what that actually meant and I can see how you came to this solution.

What I meant was creating temp files in the context of the test framework. The test framework has built in functions to create temp files that are need in the test processing... See RunTestOptions.groovy for some examples of this..

This is an example from the framework tests
```
        var source = """TESTB    START 0
         STM   14,12,12(13)
         BAS   15,8+72(15)
         ... some code..
         END   TESTB
    """
        String sourceFile = this.createTempFile("INLINE.MLC", source)
```
The test framework function `createTempFile` can be used to create files from the content of a variable. (Wondering here if Java allows binary values in strings???) 
After this code has run, the variable sourceFile has the filename for the temp file created. You then just assign the filename to an environment variable to use it.

The benefit of this is not having to manage temp files and cleanup or .gitignore to avoid committing it.
You can also add logic to vary your file contents as code... in your case replacing values with binary data where you have placed the `SKELETON>` placeholders in your skeleton source.

Having said all this, I think your code can be refactored into the test framework format so we don't have to rewrite, but reuse.. Groovy modules will accept plain old java code as well.. its backward compatible.

Might be worth having @abekornelis or @dsh33782 review the core changes as it touches all the major programs of z390.

Assuming the changes are OK, my view is we should merge this as-is and get this into the codebase. 

Once merged, I will have a go at reworking the tests into the test framework which will remove the Java program and the skeleton file. Both of these items will be part of the Groovy test script.. May take me a little bit of time so don't want to hold up the good work you have done here.

Hope this all makes sense.. and great work..

jyganci:
Anthony, all

Strongly agree that both Don and Abe should review the PR.

Are you saying that the 4 problem files -- T1#451.CPY, T1#451.DAT,
T1#451.MLC and T1P#451.MLC -- should be embedded in the framework tests?
The original PR added the 4 files to Git. The files were not handled
properly by GitHub due to the extended ASCII characters 0xA0-0xFF within
the files. Will embedding the files within the framework tests still result
in their not being properly handled by GitHub? That is, when viewing the
test containing the embedded files on GitHub, won't the files still display
incorrectly? Seeing the actual extended ASCII characters is important.

Agreed, this is a thorny issue.

Regards,

John Ganci

adelosa:
Yes. I mean embed the test files in the source for the test. 
In terms of embedding binary data, you can construct using Java formatting (\x00). 
I need to investigate creating binary files which may mean using bytes rather than strings.
I'll do some work to convert your tests to the framework but let's not hold up the PR for this to complete.

adelosa (dismissed):
I'm ok with this from test/doc perspective.. Still looking at conversion of test to use framework.. Can @abekornelis or @dsh33782 have a look as see if they are good with this change?

dsh33782 (approved):
I reviewed 10 of 12 files (2 had too large diffs?).  I did not see any errors.  I did not run tests.
I agree with approach to replace non-printable char with "." and adding option to control 
printing of non-printable characters.

jyganci:
Don,

You can click on the blue "Load diff". It will show everything.

dsh33782 (dismissed):
I finished reviewing all the files but have not independantly run tests.  I did not see any errors, but I did not verify all the different table values.

abekornelis (dismissed):
Honestly, the size and complexity are beyond me. I did run a full build and the new test script. All looks okay to me. 
The javadoc step of the build process issued 7 warnings and 4 errors. I will leave it to you whether you want to fix them as part of this pull request, or leave them for a later change.

jyganci:
How do I see the javadoc warnings and errors?

abekornelis:
> How do I see the javadoc warnings and errors?

I do not know about Linux envs. On my Windows system, I merely start the BUILD.BAT first thing it does is compose the documentation by calling BLDJVDOC.BAT. in short: the report is 'printed' on my command window. You can call BLDJVDOC.BAT from any command prompt that supports .BAT files. You may want to create an equivalent bash file.

jyganci:
Thanks. I'll create a Linux version.

jyganci:
I fixed the 4 errors (in tz390.java; 2 occurrences of @throws FileNotFoundException, 2 occurrences of @throws UnsupportedEncodingException) by importing the required files and modifying 2 "throws Exception" lines. However, I am getting 100 warnings, not 7. ???

Running against my git main (without the is451 changes) I get 100 warnings and no errors. All the warnings are "warning: no comment" due to no javadoc preceding a method. Not sure why you only see 7 warnings.

I will push the change for the errors. Fixing the warnings is not part of the pull request.

abekornelis (approved):
Fixing warnings from javadoc should not be made part of this pull request. Completely agreed on that! Thanks for all the hard work! I am approving this pull request, but I will leave it to you to do the merge. Please consider doing a squash merge to have 1 commit in the main branch, rather than all 16 from this request ...

jyganci:
I will wait until our next Zoom meeting to make sure everyone is in agreement that it is okay to merge.

PR #457, opened 2022-11-01
Fix X2C issue in DFHMDF macro

jyganci:
Remove one line from cics/DFHMDF.MAC that was used as a workaround to previous erroneous implementation of X2C that has been fixed.


adelosa:
Ok. So HLASM macros are new to me. Was the issue with x2c that it was adding extra chars at the head and the work around was to substring the value out?

jyganci:
Maybe more than you asked for.

X2C is not a macro. It is a built-in function provided by the assembler.
They are for use by the SETA, SETB and SETC expressions. So, they are
typically used in macros, but are not macros themselves. All are described
in the HLASM Language Reference manual.

The original z390 implementation of X2C converted a single byte value to
four bytes; the first three bytes were binary zeroz. The fourth byte was
the actual value. For example, X2C('F1') was, using Java notation,
0x000000F1. The DFHMDF use, X2C('6D'), produced 0x0000006D. It  needed the
value X;6D', so it extracted it with the substring of length 1 starting at
position 4. Note that within macros, substrings start at position 1, not
position 0. The X2C built-in function converts hexadecimal bytes to
character bytes. For example, X2C('F1') = '1'. The hexadecimal value F1 is
mapped to the EBCDIC character '1', which is X'F1'.

I believe that the original X2C would convert 1 to 4 characters into an
integer (32 bits). You can see this by looking at the code commented by
Fred in #438. Something like

setc_value = ""
      + (char)tz390.ebcdic_to_ascii[seta_value >>> 24]
      + (char)tz390.ebcdic_to_ascii[seta_value >>> 16 & 0xff]
      + (char)tz390.ebcdic_to_ascii[seta_value >>> 8  & 0xff]
      + (char)tz390.ebcdic_to_ascii[seta_value        & 0xff]

      ;

so
original X2C('F1') = 0x000000F1,
original X2c('F1F2') = 0x0000F1F2,
original X2C('F1F2F3') = 0x00F1F2F3,
original X2C('F1F2F3F4') = 0xF1F2F3F4, and
original X2C( F1F2F3F4F5') fails

New X2C for the above 5 values: '1', '12', '123', '1234' and '12345'. Since
'n' = X'Fn' for n = 0-9, the results are the same as the inputs, just
treated as characters.

Assemble the following program. It computes X2C('F1') and verifies that the
answer is '1'. Should end with RC=0. Look at the PRN file.

FREDX2C  CSECT
&X2C     SETC  X2C('F1')
         MNOTE 'X2C(''F1'')=&X2C'  visible in PRN
         AIF   ('&X2C' NE '1').X2ERR
         AGO   .EXIT
.X2ERR   ANOP
         MNOTE 12,'X2C ERROR'
         AGO   .EXIT
.EXIT    ANOP
         END

HLASM output looks nicer:

  Active Usings: None

  Loc  Object Code    Addr1 Addr2  Stmt   Source Statement

000000                00000 00000     1 FREDX2C  CSECT

                                      2 &X2C     SETC  X2C('F1')

                                      3          MNOTE 'X2C(''F1'')=&X2C'
 visible in PRN
                                       +X2C('F1')=1

                                      4          AIF   ('&X2C' NE
'1').X2ERR
                                      5          AGO   .EXIT

                                      6 .EXIT    ANOP

                                      7          END

Regards,

John Ganci

adelosa:
Ok. Thanks for the detailed explanation. I learned something today. So after reading the doc for X2C, the change makes the function operate like HLASM.
The following examples are provided in docs (see https://www.ibm.com/docs/en/zos/2.1.0?topic=expressions-character-valued-built-in-functions)
```
X2C('')           has value '' (null string)
X2C('F3')         has value '3'
X2C('0')          has value 'n' (EBCDIC null character)
X2C('F1F2F3F4F5') has value '12345'
X2C('000F1')      has value 'nn1'
```

adelosa (approved):
Good to merge!

PR #459, opened 2022-11-03
Z390 Groovy JVM based test framework

adelosa:
The following request moves the builds using bat/bash scripts to run tests to a Groovy based test framework.

## Rationale

There are 2 main reasons to the new framework:

1. Test scripts currently need to be maintained in 2 places - A BAT file for Windows and a Bash script for *nix. This framework replaces both with a Groovy script which runs the same way across both platforms
2. Conditions are currently limited or difficult to test using bat or bash scripts -- for example, checking that the output of an assembly listing has a particular element or error present, or output file has specific values. The new framework provides the power of Java and the JUnit test framework to perform such checks. The framework also provides convenience functions to make these activities easier when testing z390 processes.

## How to use

The full details of how you write tests can be found in z390test/readme.md file.

## What has changed

The existing tests/demos run in the build.sh/build.bat files have been converted to Groovy unit tests and are now used in place of the bash/bat scripts.   

## What has not changed

I have not removed the existing BAT/BASH scripts that run the tests. The purist in me says these should be removed and people should start using the test framework, but the pragmatist in me says that we can leave them until we start seeing adoption of the new framework. Happy to go purist if people feel this is the right direction.

Any questions, please don't hesitate to ask.


z390test/README.md
```diff
@@ -0,0 +1,243 @@
...
+If you use an IDE like IntelliJ or Eclipse, they have support for Gradle test suites.
+
+https://www.jetbrains.com/help/idea/work-with-tests-in-gradle.html
+
+
+    
+## Writing test cases
+
+The following is a basic test case that executes an assembly of a module in the z390 codebase 
+and checks the return code.
```
jyganci: Suggestion only. Since this is the reader's first exposure to the process. make this wordier. Sample: "Module TESTINS1.MLC resides in the z390 tests subdirectory. The following is a basic test case that executes an assembly of the module and checks the return code."

z390test/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+Methods available :
+* asm - Assemble only
+* asml - Assemble and link. Any non zero RC on assemble will exit with RC
+* asmlg - Assemble, link and go. Any non zero RC on assemble and link will exit with RC
+* mz390 - Assemble only
+* cblc - COBOL assemble only
+* cblclg - COBOL assemble, link and go. Any non-zero return will exist with RC
+* lz390 - Link only. Output not cleared prior to run
+* ez390 - Run only. Output not cleared prior to run
```
jyganci, line 91: Not sure of your intention here, but mz390 and ez390 are usually not used directly. Instead, use names mac and exec, which correspond to bash/mac and bash/exec. Also, not a problem, but why no cblcl?

z390test/README.md
```diff
@@ -0,0 +1,243 @@
...
+     ST    15,4(15)
+     LR    13,15
+     J
+     RETURN (14,12)
+     END   TESTB
+"""
+String sourceFilename = createTempFile("INLINE.MLC", source)
+int rc = this.asm(sourceFilename, *options)
+assert rc == 0
+```
```
jyganci: The sample program has a lot of errors. Use something simpler, such as
```
TESTBR14 CSECT
         SR    15,15
         BR    14
         END
```

z390test/README.md
```diff
@@ -0,0 +1,243 @@
...
+Use the `createTempFile` method to create a file that can be
+used in the assembly.
+
+### Set environment variables
+
+```groovy
+this.env = ['SNAPOUT': basePath('zopcheck', 'SNAPOUT.TXT')]
+```
+The env property can be set with a hashlist of environment variable that will be passed
+to the z390 calls.
```
jyganci: I assume that "hashlist" means something like a list and the enclosing square brackets are what make it a list.  Will all readers know the meaning of "hashlist"?

z390test/src/main/groovy/org/z390/test/z390Test.groovy
```diff
@@ -0,0 +1,267 @@
...
+            if (deleteFile.exists()) {
+                deleteFile.delete()
+            }
+        }
+    }
+
+    def callZ390(String asmFileExcludingExtension, String command, String... args) {
+        println("Executing ${command}: ${asmFileExcludingExtension}")
+        var cmd = ["java", "-classpath", basePath('jar', 'z390.jar'),
+                   '-Xrs', '-Xms150000K', '-Xmx150000K', command, asmFileExcludingExtension, *args].join(" ")
```
jyganci: All current bash scripts that invoke java use the z390 root directory as the "jar" directory. Look in bash/mz390. You will see the variable jardir set to a value, then the java invocation begins "java -classpath ${jardir}/z390.jar ...". The value of the jardir variable is the full path name to the z390 directory. You can see this by setting "debug-1" on bash/mz390 line 10 and running an assembly. In particular, the z390.jar used is the one in the z390 directory, not the one in the z390/jar directory. If I understand line 110 above, the generated cmd value will use the z390.jar in the z390/jar directory. This is different from normal z390 usage. The two z390.jar files should be the same, but you can't know for sure.

z390test/src/test/groovy/org/z390/test/RunAsmDemos.groovy
```diff
@@ -0,0 +1,42 @@
...
+        assert rc == 0
+    }
+    @Test
+    void test_DEMOM8Q1() {
+        int rc = this.asm(basePath("demo", "DEMOM8Q1"), *options)
+        this.printOutput()
+        assert rc == 0
+    }
+    @Test
+    void test_DEMODCB1() {
```
jyganci: "test_DEMODCB1()" should be "test_TESTDCB1()".

z390test/src/test/groovy/org/z390/test/RunAsmTests.groovy
```diff
@@ -0,0 +1,101 @@
...
+    }
+    @Test
+    void test_TESTDFP2() {
+        int rc = this.asmlg(basePath("tests", "TESTDFP2"), *options, 'optable(z390)')
+        this.printOutput()
+        assert rc == 0
+    }
+    @Test
+    void test_TESTLITS() {
+        int rc = this.asmlg(basePath("tests", "TESTLITS"), *options, 'trace', 'noloadhigh')
```
jyganci: 'trace' and 'noloadhigh' are in options.

z390test/src/test/groovy/org/z390/test/RunAsmTests.groovy
```diff
@@ -0,0 +1,101 @@
...
+    void test_TESTDC3() {
+        int rc = this.asmlg(basePath("rt", "mlc", "TESTDC3"), *options, 'optable(z390)')
+        this.printOutput()
+        assert rc == 0
+    }
+    @Test
+    void test_ZOPCHECK() {
+        var syscpyOption = "SYSCPY(${basePath('zopcheck')}+${basePath('mac')})"
+        this.env = ['SNAPOUT': basePath('zopcheck', 'SNAPOUT.TXT')]
+        int rc = this.asmlg(basePath("zopcheck", "ZOPCHECK"), *options, syscpyOption, 'optable(z390)')
```
jyganci: rt/bash/zopcheck has "bal" as an option. Add "bal" on line 97 between "*options" and "syscpyOption.

z390test/src/test/groovy/org/z390/test/RunAsmDemos.groovy
```diff
@@ -0,0 +1,42 @@
+package org.z390.test
+
+import org.junit.jupiter.api.Test
+
+class RunAsmDemos extends z390Test {
+
+    var options = ['trace', 'noloadhigh',
+                   "SYSMAC(${basePath("mac")})",
+                   "SYSOBJ(${basePath("linklib")})"]
```
jyganci: Why did you add SYSOBJ? It is not in bash/runasmdemos. Adding it will clutter the linklib directory with  .OBJ files.

I agree that not putting SYSOBJ results in .OBJ files cluttering the demo directory, but better to clutter it than the linklib directory.

z390test/src/test/groovy/org/z390/test/RunAsmTests.groovy
```diff
@@ -0,0 +1,101 @@
+package org.z390.test
+
+import org.junit.jupiter.api.Test
+
+class RunAsmTests extends z390Test {
+
+    var options = ['trace', 'noloadhigh', "SYSMAC(${basePath("mac")})", "SYSOBJ(${basePath("linklib")})"]
```
jyganci: Why did you add SYSOBJ? See RunAsmDemos.groovy comments.

z390test/src/test/groovy/org/z390/test/RunAssist.groovy
```diff
@@ -0,0 +1,48 @@
+package org.z390.test
+
+import org.junit.jupiter.api.Test
+
+
+class RunAssist extends z390Test {
+
+    var options = ['ASSIST', 'TRACEALL', "SYSMAC(${basePath("mac")})", "SYSOBJ(${basePath("linklib")})"]
```
jyganci: Same question about SYSOBJ.

z390test/src/test/groovy/org/z390/test/RunAssist.groovy
```diff
@@ -0,0 +1,48 @@
...
+    @Test
+    void test_DEMOAST1() {
+        this.env.put('XPRNT', basePath('assist', 'demo', 'DEMOAST1.XPR'))
+        int rc = this.asmlg(basePath("assist", "demo", "DEMOAST1"), *options)
+        this.printOutput()
+        assert rc == 0
+    }
+
+    @Test
+    void test_TESTINS1() {
```
jyganci: "test_TESTINS1" should be "test_TESTAST1".

z390test/src/test/groovy/org/z390/test/RunAssist.groovy
```diff
@@ -0,0 +1,48 @@
...
+        this.env.put('XPRNT', basePath('assist', 'test', 'TSTXDECI.XPR'))
+        int rc = this.asmlg(basePath("assist", "test", "TSTXDECI"), *options, 'PARM(v)')
+        this.printOutput()
+        assert rc == 0
+    }
+
+    @Test
+    void test_TSTXHEXI() {
+        this.env.put('XPRNT', basePath('assist', 'test', 'TSTXHEXI.XPR'))
+        int rc = this.asmlg(basePath("assist", "test", "TSTXHEXI"), *options, 'PARM(v)')
```
jyganci, line 43: options contains TRACEALL. bash/runassist does not use TRACEALL for TSTXDECI, TSTXHEXI.

z390test/src/test/groovy/org/z390/test/RunAssist.groovy
```diff
@@ -0,0 +1,48 @@
...
+        ]
+        int rc = this.asmlg(basePath("assist", "test", "TESTAST1"), *options)
+        this.printOutput()
+        assert rc == 0
+    }
+
+    @Test
+    void test_TSTXDECI() {
+        this.env.put('XPRNT', basePath('assist', 'test', 'TSTXDECI.XPR'))
+        int rc = this.asmlg(basePath("assist", "test", "TSTXDECI"), *options, 'PARM(v)')
```
jyganci, line 35: options contains TRACEALL. bash/runassist does not use TRACEALL for TSTXDECI, TSTXHEXI.

z390test/src/main/groovy/org/z390/test/z390Test.groovy
```diff
@@ -0,0 +1,267 @@
...
+         */
+        this.clean(cobFilename)
+        int rc
+        rc = this.callZ390(cobFilename, 'zc390', *args)
+        if (rc != 0) return rc  // exit if issue
+
+        var cobfiledir = new File(cobFilename).getParent()
+        var cobOptions = ['BAL', 'NOLISTCALL', 'MAXGBL(1500000)',
+                      "SYSMAC(${basePath('zcobol', 'mac')}+${basePath('mac')})".toString(),
+                      "SYSCPY(${cobfiledir}+${basePath('zcobol', 'cpy')})".toString()
```
jyganci, line 261: While it may be okay, you've deviated from the use of "@<path>/CBLOPT", instead putting those values here.

z390test/src/test/groovy/org/z390/test/RunCmdProc.groovy
```diff
@@ -0,0 +1,89 @@
+package org.z390.test
+
+import org.junit.jupiter.api.Test
+
+class RunCmdProc extends z390Test {
+    /*
+    Because the source contains hard coded file paths, we need to replace these with a path using basePath method.
+    We load the source into a variable and the update the file paths using a series of regex replacements
```
jyganci: Change "... and the update ..." to "... and then update ...".

This test is a tough one! Impressive work. 

After posting, I've been thinking about this one. You did a lot of work to create this test. If I wanted to change the test code (I don't), then it might take more time to update the groovy test. One way or another, either the assembler source or the groovy test has to determine whether the test is being run under Windows or Linux/Mac. I'm pretty sure I can modify the assembler source to do this at assembly time using the SYSPARM option. That way the groovy code would just need to use your hack to determine whether it's running on Windows or Linux/Mac and invoke asmlg with the appropriate SYSPARM value. I know you don't like this technique, but for this case it seems appropriate. After all, this is testing, not production. I'm going to try out my idea and will get back to you. Your thoughts?

z390test/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+To just add a single envvar without replacing the existing:
+```groovy
+this.env.put('SNAPOUT', basePath('zopcheck', 'SNAPOUT.TXT'))
+```
```
jyganci: Without replacing the existing ... what? If SNAPOUT already exists, will this env.put replace it with the new value?

z390test/src/test/groovy/org/z390/test/RunRTBSAM.groovy
```diff
@@ -0,0 +1,41 @@
...
+        rc = asmlg(basePath('bsam', "BLD${postfix}"), *options)
+        printOutput()
+        assert rc == 0
+        env.put('SYSUT1', basePath('bsam', "TEST${postfix}.TFV"))
+        rc = asmlg(basePath('bsam', "CHK${postfix}"), *options)
+        printOutput()
+        assert rc == 0
+        env = ['SNAPOUT': 'DUMMY',
+               'SYSUT1': basePath('bsam', "TEST${postfix}.TFV"),
+               'SYSUT2': basePath('bsam', "CPYR${postfix}.VES")
```
jyganci: Could this just be 
    env.put('SYSUT2', basePath('basm', "CPYR$Ppostifx}.VES")
or were the SNAPOUT and SYSUT1 values wiped out?

The bash/runrtbsam script did have to reset SYSUT1 for the final CPYR steps because of the order of the tests. However, you isolated the W and WL portions as separate tests, so the CPYR part has SYSUT1 set correctly after the BLD,CHK part.

z390test/src/test/groovy/org/z390/test/RunCmdProc.groovy
jyganci: Update. Even though I corrected a typo in the comment above, I apparently did not read the text! The problem is in the hard coded filenames. I have two additional ways for me to fix this, avoiding Groovy requiring "inside knowledge" of the assembler source.

1. This one is easy for me: change the filenames to begin with ".\" (Windows) and "./" (Linux/Mac). For TESTCMD3.MLC change them to
    TestCmdW DC    C'.\rt\test\bat\TESTCMD3.BAT',X'00'   Windows command
    TestCmdL DC    C'./rt/test/bash/testcmd3',X'00' Linux/Mac command
Similar changes for TESTCMD4.MLC.
Running bash/runcmdproc on Linux and RUNCMDPROC.BAT on Windows works with these changes.
Could you modify your test to do just these changes this and try it? Since I do not claim any knowledge of Groovy, I do not know if the test will work. If it works, then I will submit a PR for the changes. That way,the Groovy test will be simple: keep your hack to set COMSPEC and just use asmlg as in the bash runcmdproc script and the bat RUNCMDPROC.BAT file. If it does not work, the next idea may work.

2. Assuming item 1 fails, we could add a new "env" for something like you did in the runrtbsam and runrtqsam tests:
        env = ['CMD3DIR': basePathRelative('rt', 'test')]
I'm guessing here what the relative path is. I don't know if basePathRelative allows the path to be a directory, so this may not work. If this is allowed, I can modify TESTCMD3.MLC to extract this from the environment variable and modify the file name accordingly. I will need to know if the resulting environment variable ends with the file separator character.

z390test/src/test/groovy/org/z390/test/RunSortTests.groovy
```diff
@@ -0,0 +1,82 @@
...
+        this.printOutput()
+        assert rc == 0
+
+        env.put("SYSUT1", basePath('sort', 'TESTSRT1.OUT'))
+        rc = this.asmlg(basePath("sort", "TESTSRT2"), *options)
+        assert rc == 0
+
+        env.put("SYSUT2", basePath('sort', 'TESTSRT1.IN'))
+        rc = this.asmlg(basePath("sort", "TESTSRT1"), *options)
+        assert rc == 0
```
jyganci: Repeat of lines 45-47??? Delete these lines???

z390test/src/test/groovy/org/z390/test/RunAsmDemos.groovy
adelosa: OK.. no reason for that.. must have copied from some other test. Will remove.

z390test/src/test/groovy/org/z390/test/RunAsmTests.groovy
adelosa: Added bal option

z390test/src/test/groovy/org/z390/test/RunAsmTests.groovy
adelosa: Will remove

z390test/src/test/groovy/org/z390/test/RunAsmDemos.groovy
adelosa: Will fix

z390test/src/main/groovy/org/z390/test/z390Test.groovy
adelosa: OK.. learned something today.. I think in reality they will always be the same because the jar is built in the jar folder and then copied to the root but in the name of consistency, will change to the root as per the standard scripts..

z390test/README.md
adelosa: I will update this to call it a map which is the correct terminology. It is a set of key value pairs..

z390test/README.md
adelosa: Will amend as suggested. The test program is one that I would have grabbed from an existing test..

z390test/README.md
adelosa: Will updated as suggested.

z390test/src/test/groovy/org/z390/test/RunAsmTests.groovy
adelosa: It is added because TESTDC1 and TESTASC1 reference CVTTOHEX module that is assembled as part of bldlib script and the object file is part of linklib.
I can limit the scope to just the above modules as part of the link step that requires access to the object modules.

z390test/src/test/groovy/org/z390/test/RunAssist.groovy
adelosa: Will remove the option.

z390test/src/test/groovy/org/z390/test/RunAssist.groovy
adelosa: Will fix

z390test/src/test/groovy/org/z390/test/RunAssist.groovy
adelosa: Will remove TRACEALL option from common options and add to just test_DEMOAST1 and test_TESTAST1

z390test/src/test/groovy/org/z390/test/RunAssist.groovy
adelosa: Will remove TRACEALL option from common options and add to just test_DEMOAST1 and test_TESTAST1

z390test/src/main/groovy/org/z390/test/z390Test.groovy
adelosa: To be honest, the only time I have played with option files is in the tests for the option files.. I generally don't use them. I could look at changing it, but I think its simpler when all the options are in one spot.. Lets discuss this one.

z390test/src/test/groovy/org/z390/test/RunCmdProc.groovy
adelosa: I'll read this later and let you know my thoughts.. open to ideas on this because it was a PITA handling the filenames.

z390test/README.md
adelosa: Will change to "To just add a single envvar without replacing any previously set environment variables:"
And yes, if SNAPOUT already exists, this command will replace it.. 
this.env is a variable.. If you write `this.env = something` then you are fully replacing it.
If you write `this.env.put('key', 'value')` then you are adding if it does not exist and updating if it does. This is just behavior of the Map, not the framework.

z390test/src/test/groovy/org/z390/test/RunRTBSAM.groovy
adelosa: OK.. can see what you are saying.. just add the changed one in third step.. Yes.. you can do that.. but code is cheap and I almost prefer resetting the full env rather than relying on the previous steps values.. I'll change it to be consistent by using put but would also be happy redefining all envvars in each step for clarity.

z390test/src/test/groovy/org/z390/test/RunSortTests.groovy
adelosa: OK... will fix this

z390test/src/test/groovy/org/z390/test/RunCmdProc.groovy
adelosa: OK.. So I think I don't like option 1 above because it assumes that the z390 source and binary is in a specific location relative to the current directory. The Environment variable `Z390_PROJECT_ROOT` does have a default that assumes this but I don't want the tests to build this assumption in just in case someone has their source on their machine setup differently. 
For this reason, I do prefer option 2. Let the script pass the filename. This does not require special handling because the basepath and basepathrelative functions will use the correct slash depending on the OS that the test is being run with.. So no special handling would be required in the test to support both which is what we want.. 
I generally like full paths for filenames but when replacing the filename in the assembly, it was too long, so I created the basepathrelative function that provides a relative link which is shorter. I am assuming that if you go the envvar path that we would not be updating the source and that there would not be a restriction on the filename length.. 
Hope this makes sense.

adelosa:
Hi @jyganci. Thank you so much for the full review of the change. I am interested to hear on if you would feel comfortable creating tests using the framework as the intent is to replace bash/bat files for tests.

z390test/README.md
```diff
@@ -0,0 +1,239 @@
...
+var tempFileContents = """Line 1: Some text
+Line 2: This is a test file
+"""
+var tempFilename = createTempFile("myTempFile.txt", tempFileContents)
+println(tempFilename)
+```
+
+Use this to create temp files used in your tests. Returns full path to temp filename.
+
+Note - Temp directory is deleted at end of testcase.
```
jyganci: Temp directory??? Do you mean the temp file? Which brings up another question: are the temp files put in a temp directory?

z390test/src/test/groovy/org/z390/test/RunAsmDemos.groovy
```diff
@@ -0,0 +1,40 @@
...
+        assert rc == 0
+    }
+    @Test
+    void test_TESTDCB1() {
+        // Set input files
+        env.put('SYSUT1', basePath('demo', 'TESTDCB1.TF1'))
+        env.put('SYSUT2', basePath('demo', 'TESTDCB1.TF2'))
+        env.put('SYSUT3', basePath('demo', 'TESTDCB1.TF3'))
+        // run the job
+        int rc = this.asm(basePath("demo", "TESTDCB1"), *options)
```
jyganci: "this.asm" should be "this.asmlg".

z390test/src/test/groovy/org/z390/test/RunAsmDemos.groovy
```diff
@@ -0,0 +1,40 @@
...
+        // Load files to fileData
+        loadFile(basePath('demo', 'TESTDCB1.TF1'), 'TF1')
+        loadFile(basePath('demo', 'TESTDCB1.TF2'), 'TF2')
+        loadFile(basePath('demo', 'TESTDCB1.TF3'), 'TF3')
```
jyganci: TF3 not used in the check. No need to load it.

z390test/src/test/groovy/org/z390/test/RunAsmDemos.groovy
```diff
@@ -0,0 +1,40 @@
...
+        int rc = this.asm(basePath("demo", "DEMOM8Q1"), *options)
+        this.printOutput()
+        assert rc == 0
+    }
+    @Test
+    void test_TESTDCB1() {
+        // Set input files
+        env.put('SYSUT1', basePath('demo', 'TESTDCB1.TF1'))
+        env.put('SYSUT2', basePath('demo', 'TESTDCB1.TF2'))
+        env.put('SYSUT3', basePath('demo', 'TESTDCB1.TF3'))
```
jyganci: "SYSUT3" should be "SYSOUT".

z390test/src/test/groovy/org/z390/test/RunRTBSAM.groovy
```diff
@@ -0,0 +1,38 @@
+package org.z390.test
+
+import static org.junit.jupiter.api.DynamicTest.dynamicTest
+import org.junit.jupiter.api.DynamicTest
+import org.junit.jupiter.api.TestFactory
+
+class RunRTBSAM extends z390Test {
+
+    var options = ["SYSMAC(${basePath("mac")})"]
```
jyganci: Add TRACE to options.

z390test/src/test/groovy/org/z390/test/RunRTQSAM.groovy
```diff
@@ -0,0 +1,34 @@
+package org.z390.test
+
+import static org.junit.jupiter.api.DynamicTest.dynamicTest
+import org.junit.jupiter.api.DynamicTest
+import org.junit.jupiter.api.TestFactory
+
+class RunRTQSAM extends z390Test {
+
+    var options = ["SYSMAC(${basePath("mac")})", "TIME(30)"]
```
jyganci: Add TRACE to options.

z390test/src/test/groovy/org/z390/test/RunSortTests.groovy
```diff
@@ -0,0 +1,78 @@
...
+
+
+@TestInstance(Lifecycle.PER_CLASS)
+class RunSortTests extends z390Test {
+
+    var options = [
+            'noloadhigh',
+            "SYSMAC(${basePath("mac")})",
+            "SYSCPY(${basePath("mac")})",
+            "SYSOBJ(${basePath("linklib")})"
```
jyganci: Remove SYSOBJ(...) parameter.

z390test/src/test/groovy/org/z390/test/RunTestOption.groovy
```diff
@@ -0,0 +1,109 @@
...
+    @Test
+    void test_TESTOPT3L_incorrectReference() {
+        /**
+         * test 4 - should assemble with error for incorrect reference
+         */
+        int rc = this.asm(basePath("rt", "mlc", "TESTOPT"), "@${basePath("rt", "opt", "testopt3L.OPT")}")
+        this.printOutput()
+        assert rc == 16
+        assert this.stdout =~ /TZ390E abort error 21 - invalid options -\s{2}@rt\/opt\/testoptX/
+        assert this.fileData['ERR'] =~ /TZ390E abort error 21 - invalid options -\s{2}@rt\/opt\/testoptX/
```
jyganci, line 61: Not sure when testing on Windows what z390 does with the Linux path separator characters in testopt3L.opt. In any case, Windows or Linux/Mac, it does contain an invalid file name, so okay.

z390test/src/test/groovy/org/z390/test/RunTestOption.groovy
```diff
@@ -0,0 +1,109 @@
...
+
+    @Test
+    void test_TESTOPT4L_selfReference() {
+        /**
+         * test 5 - should assemble with error for self-reference
+         */
+        var OptionFileContents = """
+* Test with a self-reference
+  chksrc(2)  * Check sources for illegal characters
+@{{fullFileName}}"""
```
jyganci, line 72: Don't understand this. Enclosing double braces? What is fullFileName? I'll have to trust you on this one.

z390test/src/test/groovy/org/z390/test/TestZ390Test.groovy
```diff
@@ -0,0 +1,50 @@
...
+        var source = """TESTB    START 0
+         USING *,13
+         STM   14,12,12(13)
+         ST    13,8(13)
+         ST    15,4(15)
+         LR    13,15
+         J
+         RETURN (14,12)
+         END   TESTB
+    """
```
jyganci: While the above sample code will produce the expected error message, it contains several coding errors unrelated the test. Replace it with
```
TESTB    START 0
         STM   14,12,12(13)
         BAS   15,8+72(15)
         USING *,13
         DC    18F'0'
         ST    13,4(15)
         ST    15,8(13)
         LR    13,15
         J
         L     13,4(13)
         RETURN (14,12)
 
         END   TESTB
```
If you put an asterisk in column 1 of the "J" line, running asmlg on the file will assemble, link and run to completion without error. The existing source with the "J" line commented will assemble and link, but will abnormally terminate during execution.

z390test/src/main/groovy/org/z390/test/z390Test.groovy
```diff
@@ -0,0 +1,267 @@
...
+    @BeforeEach
+    void setUp() {
+        this.stdout = ""
+        this.stderr = ""
+        this.fileData = [:]
+    }
+    @AfterEach
+    void cleanUp() {
+        if (this.tempDir)
+            this.tempDir.deleteDir()
```
jyganci, line 43: Is the if statement merely checking that tempDir is not null?  If "yes", then the directory will be deleted but variable tempDir will still be not null, won't it? Might this cause a problem in the following createTempFile() method? Just asking.

z390test/src/main/groovy/org/z390/test/z390Test.groovy
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    def createTempFile(String fileName, String fileContents, boolean returnExt=true) {
+        /**
+         * Utility method for creating temp option files used in tests
+         */
+        if (!tempDir) {
+            this.tempDir = File.createTempDir()
+            this.tempDir.deleteOnExit()
+        }
```
jyganci, line 53: The Groovy documentation I found at https://groovy-lang.org/gdk.html does not say where the temporary directory is created. Does this matter?

z390test/src/main/groovy/org/z390/test/z390Test.groovy
```diff
@@ -0,0 +1,267 @@
...
+        var fullFileName = pathJoin(tempDir.absolutePath, fileName)
```
jyganci, line 54: Where is the absolutePath field defined? I don't see anything in the Java documentation or the Groovy documentation for File.

z390test/src/main/groovy/org/z390/test/z390Test.groovy
```diff
@@ -0,0 +1,267 @@
...
+        println("Creating temp source: ${fullFileName}")
```
jyganci, line 55: Eventually delete this line???

z390test/src/main/groovy/org/z390/test/z390Test.groovy
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    String basePathRelative(String... pathItems) {
+        var fullPathItems = [this.project_root, *pathItems]
+        return fullPathItems.join(File.separator)
+    }
+
+    static String filenameWithoutExtension(String filename) {
+        String result = filename[0..<filename.lastIndexOf('.')]
+        println(result)
```
jyganci: Eventually delete this line???

z390test/src/test/groovy/org/z390/test/RunAsmTests.groovy
adelosa: The commit now limits the obj option to those steps that require access to CVTTOHEX obj.

z390test/README.md
adelosa: I do mean directory, but thats because the temp file is generated in a temp directory -- at least on Windows and MacOS.. I think Linux may just use the /tmp folder.. I will update to say that the temp file and any generated directory containing the temp file will be deleted.

z390test/src/test/groovy/org/z390/test/RunAsmDemos.groovy
adelosa: Will fix this.. good pickup

z390test/src/test/groovy/org/z390/test/RunAsmDemos.groovy
adelosa: Can remove, but should it also be removed from the input or is this still required?
Line 26 in RunAsmDemos.groovy
```groovy
        env.put('SYSUT3', basePath('demo', 'TESTDCB1.TF3'))
```

adelosa: Will load as SYSOUT file for debugging purposes and can be referenced in tests.. see comment on SYSOUT

z390test/src/test/groovy/org/z390/test/RunAsmDemos.groovy
adelosa: Will fix this..

z390test/src/test/groovy/org/z390/test/RunRTBSAM.groovy
adelosa: Will add TRACE option

z390test/src/test/groovy/org/z390/test/RunRTQSAM.groovy
adelosa: Will add trace

z390test/src/test/groovy/org/z390/test/RunSortTests.groovy
adelosa: Will fix

z390test/src/test/groovy/org/z390/test/RunTestOption.groovy
adelosa: The test checks that the error message matches.. I have run the tests on both Linux and Windows.. In this case, looks like a hardcoded backslash (\) is used, but as you say the test is an invalid filename so may not matter.

z390test/src/test/groovy/org/z390/test/RunTestOption.groovy
adelosa: OK.. this was one of the challenges of this test.. I needed to self reference the temp filename.. Its a chicken and egg with the temp file function.. you don't know the file name until its created, but the file contents contain the generated filename.. The function provides the special {{fullFileName}} placeholder so that the function inserts the name of the temp file at the location of {{fullFileName}}.
I have added a description of this behaviour in the README.md file for the createTempFile function:

  If you need to include the name of the generated file in the file, use `{{fullFileName}}` placeholder in
  file contents and the function will replace this with the actual filename created.

z390test/src/test/groovy/org/z390/test/TestZ390Test.groovy
adelosa: Have updated the assembler code as suggested. Generates the required error message with the J line included.

z390test/src/main/groovy/org/z390/test/z390Test.groovy
adelosa: The first time a temp file is needed in a test class using the createTempFile method, a directory name assigned to the tempFile class variable and the directory created.
```groovy
if (!tempDir) {
  this.tempDir = File.createTempDir()
  this.tempDir.deleteOnExit()
}
```
At completion of the first test that created a temp file, the directory is deleted in the cleanup method, but the tempDir is still pointing to the dir name used.
```groovy
    @AfterEach
    void cleanUp() {
        if (this.tempDir)
            this.tempDir.deleteDir()
    }
```
tempDir now is not null so a new directory or name is not created on second and subsequent calls to create temp files.. But the method that creates the file will use the directory name and presumably create the directory if it does not exist.
```groovy
        new File(fullFileName).with {
            createNewFile()
            write(fileContents)
        }
```
This is what I guess is happening, but would need to spend some time to prove this is the case..
Its a bit sloppy and could probably be done a few different ways but it does seem to be working so
if it ain't broke, don't fix it...

z390test/src/main/groovy/org/z390/test/z390Test.groovy
adelosa: No..  the location of the temp dir is OS dependant.. Linux will use /tmp, Windows uses a location in the user profile and MacOS has some special place that I can't recall at the moment.  Should not matter because the tempDir has a reference to it and can clean up after itself.

z390test/src/main/groovy/org/z390/test/z390Test.groovy
adelosa: this.tempDir is a Java.io.file type.  `this.tempDir = File.createTempDir()`
Java.io.file has a method called getAbsolutePath
See https://docs.oracle.com/javase/8/docs/api/java/io/File.html#getAbsolutePath--
In groovy, you can shortcut the way you call a getter method..
`tempDir.getAbsolutePath()`
can be also called as 
`tempDir.AbsolutePath`

z390test/src/main/groovy/org/z390/test/z390Test.groovy
adelosa: We could.. its nice for debugging to know where the temp file is created..

z390test/src/main/groovy/org/z390/test/z390Test.groovy
adelosa: I will delete this.. must have just been me debugging..

z390test/README.md
```diff
@@ -0,0 +1,242 @@
...
+Returns:
+* rc - The result code from the execution
+
+Methods available :
+* asm - Assemble only
+* asml - Assemble and link. Any non zero RC on assemble will exit with RC
+* asmlg - Assemble, link and go. Any non zero RC on assemble and link will exit with RC
+* mz390 - Assemble only
+* cblc - COBOL assemble only
+* cblclg - COBOL assemble, link and go. Any non-zero return will exist with RC
```
jyganci: For both cblc and cblclg, "assemble" should be "compile".

z390test/src/main/groovy/org/z390/test/z390Test.groovy
```diff
@@ -0,0 +1,266 @@
...
+    void setUp() {
+        this.stdout = ""
+        this.stderr = ""
+        this.fileData = [:]
+    }
+        @AfterEach
+        void cleanUp() {
+            if (this.tempDir)
+                this.tempDir.deleteDir()
+        }
```
jyganci: Why did you indent the "@AfterEach..." 4 spaces?

z390test/src/main/groovy/org/z390/test/z390Test.groovy
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    def cblc(Map kwargs=[:], String cobFilename, String... args) {
+        /**
+         * cobol compile
+         */
+        this.clean(cobFilename)
+        int rc
+        rc = this.callZ390(cobFilename, 'zc390', *args)
+        if (rc != 0) return rc  // exit if issue
```
jyganci, line 238: The bash/cblc and bat\cblc files treat rc=4 as a warning and continue with the compile.
I suspect that exiting if rc is not 0 is okay here. Just pointing out the difference.

z390test/src/test/groovy/org/z390/test/RunSortTests.groovy
```diff
@@ -0,0 +1,76 @@
...
+import org.junit.jupiter.api.TestInstance.Lifecycle
+
+
+@TestInstance(Lifecycle.PER_CLASS)
+class RunSortTests extends z390Test {
+
+    var options = [
+            'noloadhigh',
+            "SYSMAC(${basePath("mac")})",
+            "SYSCPY(${basePath("mac")})",
```
jyganci: Remove trailing comma???

z390test/src/test/groovy/org/z390/test/RunSortTests.groovy
```diff
@@ -0,0 +1,76 @@
...
+    ]
+
+    var runOptions = [
+            "SYS390(${basePath('sort')})",
```
jyganci: Remove trailing comma???

z390test/src/main/groovy/org/z390/test/z390Test.groovy
```diff
@@ -0,0 +1,266 @@
...
+        def envList = []
+        this.env.each{item -> envList.add("${item.key}=${item.value}")}
+        return envList
+    }
+
+    def clean(String asmFileExcludingExtension) {
+        /**
+         * clear files for module
+         */
+        for (ext in ["PRN", "ERR", "OBJ", "390", "OBJ", "LOG", "LST", "STA"]) {
```
jyganci, line 106: The bash/asm script also deletes (trace) files with extension TR* (that is, TRA, TRE, etc).

z390test/src/test/groovy/org/z390/test/RunRtTests.groovy
```diff
@@ -0,0 +1,101 @@
+package org.z390.test
+
+import org.junit.jupiter.api.Test
+
+class RunRtTests extends z390Test {
+    var sysmac = basePath("mac")
+    var syscpy = basePath("mac")
+    var libs = ["SYSMAC(${sysmac}+)", "SYSCPY(${syscpy}+)"]
```
jyganci: The bash/asm script has the plus sign preceding the value for both SYSMAC and SYSCPY.
So, replace above values with "SYSMAC(+${sysmac})" and "SYSCPY(+${syscpy})".

z390test/src/test/groovy/org/z390/test/RunRtTests.groovy
```diff
@@ -0,0 +1,101 @@
...
+    // rt1 = asmlg
+    var rt1Options = ['bal', 'notiming', 'stats', *libs]
+    // rt3 = mz390
+    var rt3Options = ['noasm', 'bal', 'notiming', 'stats', *libs]
+    // rt4 = mz390
+    var rt4Options = ['noasm',  'bal', 'notiming', 'stats', *libs]
```
jyganci: The rt\RT4.BAT file does not have 'noasm' --- only has 'bal', 'notiming' and 'stats'.

z390test/src/test/groovy/org/z390/test/RunRtTests.groovy
```diff
@@ -0,0 +1,101 @@
...
+    // rt7 = asml + ez390
+    var rt7AsmlOptions = ['bal', 'notiming', *libs]
+    var rt7Ez390Options = ['notiming', 'stats', *libs]
+
+    @Test
+    void test_TESTERR1() {
+        // rt3 = mz390
+        var syscpyRtTest = "SYSCPY(${basePath('rt', 'test')})"
+        int rc = this.mz390(basePath("rt", "test", "TESTERR1"), *rt3Options, syscpyRtTest)
```
jyganci: Why did you add the syscpyRtTest value?

z390test/src/test/groovy/org/z390/test/RunRtTests.groovy
```diff
@@ -0,0 +1,101 @@
...
+    @Test
+    void test_TESTERR7() {
+        // rt7 = asml + mz390
+        this.env = ['TESTERR7': 'afile.txt']
+        int rc
+        rc = this.asml(basePath("rt", "test", "TESTERR7"), *rt7AsmlOptions)
+        assert rc == 0
+        rc = this.ez390(basePath("rt", "test", "TESTERR7"), *rt7Ez390Options)
+        this.printOutput()
+        assert rc == 16
```
jyganci, line 77: While not necessary, you could add an assert statement as you did for TESTERR4 mentioning
the S013 abend.

z390test/src/main/groovy/org/z390/test/z390Test.groovy
adelosa: No idea.. will fix that

z390test/src/main/groovy/org/z390/test/z390Test.groovy
adelosa: It should be aligned to the bash/bat scripts so happy to update.
Will change to exit if not 0 or 4.

z390test/src/test/groovy/org/z390/test/RunSortTests.groovy
adelosa: OK, will remove.. although given it is valid in Groovy I often leave them in in case I need to add additional entries.. Python has the same behaviour.

z390test/src/test/groovy/org/z390/test/RunSortTests.groovy
adelosa: ok, will fix

z390test/src/main/groovy/org/z390/test/z390Test.groovy
adelosa: OK, I have added logic to the class to support deleting all TR* files as part of the clean process.

z390test/src/test/groovy/org/z390/test/RunRtTests.groovy
adelosa: Will update this

z390test/src/test/groovy/org/z390/test/RunRtTests.groovy
adelosa: Will update this

z390test/src/test/groovy/org/z390/test/RunRtTests.groovy
adelosa: Not sure, may have been a cut and paste issue.. have removed and test still runs.

z390test/src/test/groovy/org/z390/test/RunRtTests.groovy
adelosa: Added some additional asserts

z390test/src/main/groovy/org/z390/test/z390Test.groovy
adelosa: Are you OK if I leave this as is or do you want me to change to use the option files?

z390test/src/test/groovy/org/z390/test/RunCmdProc.groovy
adelosa: @jyganci Lets work together on a way to fix this one.. would love to remove the process of copying and modifying the source to get the tests to run..

z390test/src/main/groovy/org/z390/test/z390Test.groovy
adelosa: @jyganci -- let me know your thoughts on this and if I need to make some more changes

jyganci: My only concern is the use of the tempDir variable after the directory is deleted.
I am assuming that you've tested the reuse of tempDir after a deletion of the
directory. If so, then I'll agree that there is not an issue here.

z390test/src/main/groovy/org/z390/test/z390Test.groovy
```diff
@@ -0,0 +1,274 @@
...
+        /**
+         * clear files for module
+         */
+        var filename_only = new File(asmFileExcludingExtension).name
+        var regex = "${filename_only}\\.TR."
+        var dirName = new File(asmFileExcludingExtension).getParent().toString()
+        var traceFiles = new FileNameByRegexFinder().getFileNames(dirName, regex)
+        for (filename in traceFiles) {
+            new File(filename).delete()
+        }
```
jyganci, line 104: Nice!

z390test/src/main/groovy/org/z390/test/z390Test.groovy
jyganci: Okay to leave as is.

z390test/src/test/groovy/org/z390/test/RunRtTests.groovy
```diff
@@ -0,0 +1,104 @@
...
+    }
+    @Test
+    void test_TESTERRA() {
+        // rt4 = mz390
+        int rc = this.mz390(basePath("rt", "test", "TESTERRA"), 'CHKMAC(2)', *rt4Options)
+        this.printOutput()
+        assert rc == 16
+    }
+}
```
jyganci: Add newline at end of file???

jyganci:
Anthony,

I agree. As stated in an earlier email, I've modified the TESTCMD3.MLC and
TESTCMD4.MLC files to
1. get the "prefix" (directory name) where the test files reside from an
environment variable
2. determine whether Windows or Linux/Mac via PARM(W) or PARM(L).

In addition, I've modified the bash/runcmdproc script to
3. export the "prefix" value to an environment variable
4. invoke TESTCMD3 with PARM(L)
5. invoke TESTCMD4 with PARM(L,x) where x=1.2.3

I believe that with these changes, the Groovy code will be a lot like that
for the SORT test (or any test that exports file names).

Rather than email back and forth, how about a face-to-face? Zoom? Whereby?
Facetime?

Regards,

John Ganci

z390test/src/test/groovy/org/z390/test/RunCmdProc.groovy
adelosa: @jyganci -- after our discussion, I think the decision was to leave this as is and depending on who goes first in terms of merges, we will deal with the update to pass the file names to the test modules via envvars. Let me know if I got this correct

z390test/src/test/groovy/org/z390/test/RunRtTests.groovy
adelosa: Added newline in fde2a52

adelosa:
@jyganci - I assume you are good to approve this. If not, let me know what else we need.

jyganci (approved):

PR #471, opened 2023-02-13
remove readme.txt that has CR,LF line endings

jyganci:
Removing problem text file that has Windows line endings. After this is merged, will add the file, with *nix line endings,


adelosa (approved):
Lets move forward with this.

PR #474, opened 2023-02-26
fetch upstream when git describe error during build

### LINKED ISSUE #472: Build failing when no repo tags
As explained, the break was due to a lack of wifi at my aged-care facility.
A friend managed to set me up with wifi via my phone.

cd c:\z390dev\z390

entered ..... build.bat
got
c:\z390dev\z390>build.bat
Could Not Find c:\z390dev\z390\jar\z390.jar
src\az390.java
src\ez390.java
src\gz390.java
src\gz390_screen.java
src\iz390.java
src\lz390.java
src\MessageBox.java
src\mz390.java
src\pz390.java
src\sz390.java
src\test_screen.java
src\test_unicode.java
src\tz390.java
src\vz390.java
src\z390.java
src\zACB.java
src\zc390.java
src\zCatVAIX.java
src\zCatVCLR.java
       19 file(s) copied.
        1 file(s) copied.
java version "17.0.1" 2021-10-19 LTS
Java(TM) SE Runtime Environment (build 17.0.1+12-LTS-39)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.1+12-LTS-39, mixed mode, sharing)
gz390.java:574: warning: [removal] SecurityManager in java.lang has been deprecated and marked for removal
            SecurityManager sm = System.getSecurityManager();
            ^
gz390.java:574: warning: [removal] getSecurityManager() in System has been deprecated and marked for removal
            SecurityManager sm = System.getSecurityManager();
                                       ^
z390.java:423: warning: [removal] SecurityManager in java.lang has been deprecated and marked for removal
            SecurityManager sm = System.getSecurityManager();
            ^
z390.java:423: warning: [removal] getSecurityManager() in System has been deprecated and marked for removal
            SecurityManager sm = System.getSecurityManager();
                                       ^
4 warnings
fatal: No names found, cannot describe anything.
Version is uncontrolled - add FORCE parm to ignore
bat\bldjar ERROR: Error finding z390 version number
build.bat ERROR: Encountered RC 8 - exiting

c:\z390dev\z390>

John Hennesy

adelosa:
This request is to assist new contributors to the z390 project.

Fixes #472

The flow for a new contributor is that they would fork the main repository in GitHub and then clone the fork to their local machine.
What we have discovered is that the GitHub fork process does not bring across the tags from the main repo, and this extends to the local copy. This lack of tags causes a build issue. 

When a build is executed, it creates a z390.properties file which contains the current version which is sourced from the git repo tags and is included in the jar. The properties file is generated by the relver script (bash/relvel, bat/RELVER.BAT) 
Because the clone does not have any tags so the git describe action fails and you cannot override by using FORCE parameter.

Solution:
The relver script has been modified to fetch upstream tags when a git describe error is encountered.
If not present, the script will add the upstream remote required to fetch the tags. 
I have tested this against a new fork which does not have the tags present (chookperson/z390)

@abekornelis - are you able to update bat/RELVER.BAT to do the equivalent function for Windows. I don't have any Windows machines I can build this with. Also let me know if you are OK with the approach taken.

### DIFF
--- a/bash/relver
+++ b/bash/relver
@@ -9,6 +9,13 @@ cd ..
 
 VER="Uncontrolled";
 
+get_tags() {
+    if ! git config remote.upstream.url > /dev/null; then
+        git remote add upstream https://github.com/z390development/z390.git
+    fi
+    git fetch -p upstream
+}
+
 # is git installed
 $(git --version > /dev/null 2>&1)
 GIT_INSTALLED=$?
@@ -21,6 +28,12 @@ else
     if [ ${GIT_REPO} != 0 ]; then
         echo >&2 "Not a git repo.";
     else
+        $(git describe > /dev/null 2>&1)
+        GIT_DESCRIBE=$?
+        if [ ${GIT_DESCRIBE} != 0 ]; then
+            echo >&2 "Unable to git describe - fetching tags from upstream"
+            get_tags
+        fi
         VER=$(git describe);
     fi    
 fi
--- a/bat/RELVER.BAT
+++ b/bat/RELVER.BAT
@@ -13,16 +13,49 @@ rem ----- Lvl(%z_NestLevel%) Start %0 %1 %2 %3 %4 %5 %6 %7 %8 %9
 pushd %~dps0..
 set z_Version="Uncontrolled"
 
-git --man-path >NUL
-
+rem -- is git installed?
+git --version >NUL
 set z_Git_Installed=%ERRORLEVEL%
 if %z_Git_Installed% NEQ 0 (echo Git is not installed
                             goto final
-                            ) 
+                            )
 
+rem -- are we in a git repository?
+git rev-parse --is-inside-work-tree >NUL
+set z_Git_WorkTree=%ERRORLEVEL%
+if %z_Git_WorkTree% NEQ 0 (echo Git cannot determine whether we're in a work tree
+                           goto final
+                           )
 for /F "tokens=* USEBACKQ" %%F in (`git rev-parse --is-inside-work-tree`) do (set z_Git_Repo=%%F)
 if "%z_Git_Repo%" NEQ "true" (echo Not a git repository
                               goto final)
+
+rem -- get git to describe the repo
+git describe >NUL
+set z_Git_Describe=%ERRORLEVEL%
+if %z_Git_Describe% EQU 0 goto describe
+echo Unable to git describe - fetching tags from upstream
+
+rem -- make sure we have an upstream defined
+git config remote.upstream.url >NUL
+set z_Git_Remote=%ERRORLEVEL%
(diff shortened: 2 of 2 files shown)


abekornelis:
Anthony, thanks for the quick fix. I definitely agree with the approach you have chosen.
However, to double-check your changes I'll need some more time - I'm not sure I'll be able to pull it off by myself.
I'll give it a try when I find some time. FYI: John seems to have re-installed some windoze stuff. He may be able to help you a lot faster than I ever could hope to ...

adelosa:
Hi @abekornelis - I updated bat/RELVER.BAT with the fix that you provided. Hopefully I got it correctly. Setting this to ready for merge. Will see if @jyganci can review since we have both contributed to this request.

abekornelis (approved):
looks good to me. I'll leave it to John to speak the final word on this request.

bash/relver
```diff
@@ -9,6 +9,13 @@ cd ..
 
 VER="Uncontrolled";
 
+get_tags() {
+    if ! git config remote.upstream.url > /dev/null; then
```
jyganci, line 13: Not sure about this. Does the condition need to be enclosed by brackets? That is,
    if [ ! git config remote.upstream.url > /dev/null; ] then

Also, add 2>&1 at the end?

bat/RELVER.BAT
```diff
@@ -13,16 +13,49 @@ rem ----- Lvl(%z_NestLevel%) Start %0 %1 %2 %3 %4 %5 %6 %7 %8 %9
...
+rem -- is git installed?
+git --version >NUL
 set z_Git_Installed=%ERRORLEVEL%
 if %z_Git_Installed% NEQ 0 (echo Git is not installed
                             goto final
-                            ) 
+                            )
 
+rem -- are we in a git repository?
+git rev-parse --is-inside-work-tree >NUL
```
jyganci, line 24: Suggestion only. The bash version added 2?&1 on the end of this command. You might do the same. There are a few other places this was done as well.

jyganci (approved):
Made a suggestion. Make changes if you want.

bat/RELVER.BAT
adelosa: @jyganci  - I am not sure that sysout and syserr redirection works the same way in Windows.. This is a bash construct, not Git. See https://linuxize.com/post/bash-redirect-stderr-stdout/ I am sure that @abekornelis can confirm.

abekornelis: That might work - but I am loath to suppress error info that might come in handy in case a problem surfaces. Redirecting the 'standard output' is no problem as that would be - in this case - only useless clutter on the screen.

jyganci:
Okay to both replies. Merge the PR as-is.

PR #475, opened 2023-02-27
Is470: Add Linux, MacOS support to zVSAM versions 1 and 2; enhance and fix existing Windows support

jyganci:


abekornelis (approved):
John, I did a repeat test of all vsam1\test\bat\runtests.bat and vsam1\demo\bat\rundemos.bat
and also vsam2\bat\runtests.bat
I have noticed no problems - you have my go.

jyganci:
Abe,

Thanks. Before I do the merge, a couple of things.

1. Did you look at any of the bat files or hash scripts? Yes, I know there
are a lot of them, but they are all pretty much the same. Modeled after
Melvyn's cics/bat files and my cics/bash scripts. If you would. please take
a quick look at one or two pairs, such as

vsam1/demo/bash/esf1demo vs vsam1/demo/bat/ESF1DEMO.BAT and
vsam1/demo/bash/ksf1rpo1 vs vsam1/demo/bat/KSF1RPO1.BAT

Are you okay with the style?

2. If you recall, I did get one abend running vsam1/demo/bash/ksf1rpo2 and
vsam1/dem//bat/KSF1RPO2.BAT when attempting to unload NAMELIST, which is a
PATH for the KSF1NAME KSDS. Unlike the other unloads of PATHs, namely
ADDRLIST, CITYLIST, STATLIST, and ZIPLIST -- which are all alternate
indexes -- NAMELIST is not an alternate index. As I've previously
mentioned, I think this is due to a problem with the zVSAM Java code, most
likely in vz390.java. I plan to look more at this, but may have questions
for you. I will also create another issue to track it.

I'll do the merge when I hear back from you.

Regards,

John Ganci

abekornelis:
John,

I did have a short look at the bat files, I'm no good at bash, so I skipped those.
The style of the bat files could be slightly different - but it's definitely good enough for now.
I'll seek to address additional improvements later, as I'd also like to include these tests in the RT suite.
I suggest you go ahead and do a merge - squash merge would be my favorite.

Abe
===

PR #478, opened 2023-03-19
redoing all changes for enabling javadoc comments

abekornelis:
Since the original branch for javadoc-enbalement had issues, I deleted it, which closed the associated merge request.
The I created a fresh branch with the combined changes from the original branch, committed, pushed, and created merge request.

### DIFF
--- a/bat/BLDJAR.BAT
+++ b/bat/BLDJAR.BAT
@@ -1,5 +1,6 @@
 @if /I "%1" == "tron" (echo on) else (echo off)
 rem create z390 jar file from sources
+rem unless NoDocs is specified as an argument, generate javadoc files
 
 setlocal
 if /I "%1" == "tron" (set z_TraceMode=tron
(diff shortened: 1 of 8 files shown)


bat/BLDJAR.BAT
```diff
@@ -12,10 +12,18 @@ set /A z_NestLevel=%z_NestLevel%+1
...
-if not exist ".\jar\" mkdir jar
-
 rem both the compiler javac.exe and jar builder jar.exe must be on sdk path
 rem note you can use current z390.jar with compatible java runtime with sdk
+
+rem first generate javadoc
+if not exist ".\docenv\javadoc\" mkdir docenv\javadoc
+cd docenv\javadoc
+call javadoc.exe ..\..\src\*.java
+cd ..\..
```
jyganci: Suggestion. Do the javadoc generation in a separate bat file. Someone may want to run BLDJAR.BAT to re-build the z390.jar file. Why should they also re-build the javadoc?

jyganci (dismissed):
Made one suggestion in BLDJAR.BAT to move javadoc creation to separate bat file that would be invoked by BUILD.BAT. Otherwise okay. Approved.

bat/BLDJAR.BAT
abekornelis: I understand your line of reasoning. However, in my mind, it makes sense to generate the doc whenever the code is compiled. Main reasons being: the doc is generated from the very same source that is being compiled. Generating the doc for every compile ensures the doc is always up to date and in sync with the current jar. And since it takes only a few seconds the overhead is manageable. 
I suggest to put the javadoc logic in a separate procedure, which will be called by default when compiling the jar. I'll add a parm to the BLDJAR.BAT so users have an option to suppress the javadoc step. I'll let it default to not suppress the javadoc since in the github logic after every push we do want to generate the latest and greatest.
Would that be okay with you?

jyganci:
It's okay to leave it as you wrote it.

jyganci (approved):

PR #484, opened 2023-05-02
Add op_type_otype table to fix O attribute value for each type of mne…

abekornelis:
Add op_type_otype table to fix O attribute value for each type of mnemonic

### DIFF
--- a/rt/mlc/ATTRIB$O.MLC
+++ b/rt/mlc/ATTRIB$O.MLC
@@ -20,17 +20,141 @@ ATTRIB$O RMODE 24
          MEND
 
 TESTOT1  CSECT
-         SHOWOP PUSH
-         SHOWOP BCR
-         SHOWOP BER
-         SHOWOP BC
-         SHOWOP BE
-         SHOWOP JE
-         SHOWOP BRC
-         SHOWOP SRNM
-         SHOWOP SAVE
-         SHOWOP XXXXX
-
+*
+* Validate O attribute for some sample menmonics:
+* A Assembler operation code
+* E Extended mnemonic operation code
+* M Macro definition O Machine operation code
+* S Macro definition found in library
+* U Undefined, unknown, unassigned, or deleted operation code
+*
+*        SHOWOP PUSH                   * A
+*        SHOWOP BCR                    * O
+*        SHOWOP BER                    * E
+*        SHOWOP BC                     * O
+*        SHOWOP BE                     * E
+*        SHOWOP JE                     * E
+*        SHOWOP BRC                    * O
+*        SHOWOP SRNM                   * O
+*        SHOWOP SHOWOP                 * M
+*        SHOWOP SAVE                   * S
+*        SHOWOP XXXXX                  * U
+*
          XR    15,15                   * Set RC=0
          BR    14                      * Return
+*
+* Validate assembly of mnemonics with changed definitions
+*
+CLT      CLT      1,B'1010',7(5)       * EB1A 5007 0023
+         CLTH     1,7(5)               * EB12 5007 0023
+         CLTL     1,7(5)               * EB14 5007 0023
+         CLTNE    1,7(5)               * EB16 5007 0023
+         CLTE     1,7(5)               * EB18 5007 0023
+         CLTNL    1,7(5)               * EB1A 5007 0023
+         CLTNH    1,7(5)               * EB1C 5007 0023
+*
+CLGT     CLGT     1,B'1010',7(5)       * EB1A 5007 002B
+         CLGTH    1,7(5)               * EB12 5007 002B
+         CLGTL    1,7(5)               * EB14 5007 002B
+         CLGTNE   1,7(5)               * EB16 5007 002B
+         CLGTE    1,7(5)               * EB18 5007 002B
+         CLGTNL   1,7(5)               * EB1A 5007 002B
+         CLGTNH   1,7(5)               * EB1C 5007 002B
+*
+LOCFHR   LOCFHR   1,2,B'1010'          * B9E0 A012
+         LOCFHRO  1,2                  * B9E0 1012
+         LOCFHRH  1,2                  * B9E0 2012
+         LOCFHRP  1,2                  * B9E0 2012
+         LOCFHRL  1,2                  * B9E0 4012
+         LOCFHRM  1,2                  * B9E0 4012
+         LOCFHRNE 1,2                  * B9E0 7012
+         LOCFHRNZ 1,2                  * B9E0 7012
+         LOCFHRE  1,2                  * B9E0 8012
+         LOCFHRZ  1,2                  * B9E0 8012
+         LOCFHRNL 1,2                  * B9E0 B012
+         LOCFHRNM 1,2                  * B9E0 B012
+         LOCFHRNH 1,2                  * B9E0 D012
+         LOCFHRNP 1,2                  * B9E0 D012
+         LOCFHRNO 1,2                  * B9E0 E012
+*
+LOCGR    LOCGR    1,2,B'1010'          * B9E2 A012
+         LOCGRO   1,2                  * B9E2 1012
+         LOCGRH   1,2                  * B9E2 2012
+         LOCGRP   1,2                  * B9E2 2012
+         LOCGRL   1,2                  * B9E2 4012
+         LOCGRM   1,2                  * B9E2 4012
+         LOCGRNE  1,2                  * B9E2 7012
+         LOCGRNZ  1,2                  * B9E2 7012
+         LOCGRE   1,2                  * B9E2 8012
+         LOCGRZ   1,2                  * B9E2 8012
+         LOCGRNL  1,2                  * B9E2 B012
+         LOCGRNM  1,2                  * B9E2 B012
+         LOCGRNH  1,2                  * B9E2 D012
+         LOCGRNP  1,2                  * B9E2 D012
+         LOCGRNO  1,2                  * B9E2 E012
+*
+LOCR     LOCR     1,2,B'1010'          * B9F2 A012
+         LOCRO    1,2                  * B9F2 1012
+         LOCRH    1,2                  * B9F2 2012
+         LOCRP    1,2                  * B9F2 2012
+         LOCRL    1,2                  * B9F2 4012
+         LOCRM    1,2                  * B9F2 4012
+         LOCRNE   1,2                  * B9F2 7012
+         LOCRNZ   1,2                  * B9F2 7012
+         LOCRE    1,2                  * B9F2 8012
+         LOCRZ    1,2                  * B9F2 8012
+         LOCRNL   1,2                  * B9F2 B012
+         LOCRNM   1,2                  * B9F2 B012
+         LOCRNH   1,2                  * B9F2 D012
+         LOCRNP   1,2                  * B9F2 D012
+         LOCRNO   1,2                  * B9F2 E012
+*
+SELFHR   SELFHR   1,2,3,B'1010'        * B9C0 3A12
+         SELFHRO  1,2,3                * B9C0 3112
+         SELFHRH  1,2,3                * B9C0 3212
+         SELFHRP  1,2,3                * B9C0 3212
+         SELFHRL  1,2,3                * B9C0 3412
+         SELFHRM  1,2,3                * B9C0 3412
+         SELFHRNE 1,2,3                * B9C0 3712
+         SELFHRNZ 1,2,3                * B9C0 3712
+         SELFHRE  1,2,3                * B9C0 3812
+         SELFHRZ  1,2,3                * B9C0 3812
+         SELFHRNL 1,2,3                * B9C0 3B12
+         SELFHRNM 1,2,3                * B9C0 3B12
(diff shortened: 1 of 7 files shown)


abekornelis:
I have fixed the definitions for non-vector instructions (vector instructions to be fixed in the next change). I have also adjusted the logic to evaluate an O attribute reference. I have added an RT program, but that still needs to be included in the automated RT tests. Some help on how to achieve that would be highly appreciated.

rt/mlc/ATTRIB$O.MLC
adelosa, line 1: @abekornelis -- What is the intention of this test.. ASM only vs ASMLG. 
Are you wanting to check that the opcodes generated are the same as in the comments?
For example..
```
00099  000004                                     (1/70)109 *
00100  000004                                     (1/71)110 * Validate assembly of mnemonics with changed definitions
00101  000004                                     (1/72)111 *
00102  000004 EB1A50070023                        (1/73)112 CLT      CLT      1,B'1010',7(5)       * EB1A 5007 0023
00103  00000A EB1250070023                        (1/74)113          CLTH     1,7(5)               * EB12 5007 0023
00104  000010 EB1450070023                        (1/75)114          CLTL     1,7(5)               * EB14 5007 0023
00105  000016 EB1650070023                        (1/76)115          CLTNE    1,7(5)               * EB16 5007 0023
00106  00001C EB1850070023                        (1/77)116          CLTE     1,7(5)               * EB18 5007 0023
00107  000022 EB1A50070023                        (1/78)117          CLTNL    1,7(5)               * EB1A 5007 0023
00108  000028 EB1C50070023                        (1/79)118          CLTNH    1,7(5)               * EB1C 5007 0023
```
Instruction CLT should generate `EB1A 5007 0023` opcode

I can add checks for this in the test if that is the goal.
Let me know if there are other things that need to be validated as part of the test.

adelosa: I have pushed some changes to add your tests to the automation.
The only change to your test module was to add a OPC= and remove spaces from the expected OpCodes at the end of the lines. The test module inspects the PRN and checks expected opcodes are output based on finding OPC= on the assembly line. Hope that makes sense.

abekornelis: Anthony,

thanks for doing the hard work. From my perspective, ASM should be good enough.
the comments indeed indicate the object code to be generated.
Before that section there are a few blocks where the O attribute evaluation itself is being tested.
Likewise, expected output is defined in the comment area. Does that require additional changes to get tested automagically?

adelosa: Thats a little more tricky as the result is in the macro expansion but I will see what I can do.

abekornelis: Alternatively, and probably a lot easier, would then be to change the macro SHOWOP from reporting the O attribute value to validating the O attribute value against an additional parameter (as opposed to validating against a comment)
We could then issue an MNOTE with warning level 8 if a mismatch occurs, failing the entire assembly. If you agree with this, I'll modify the macro as needed.

adelosa: I'm adding some additional tests to check macro expansion for SHOWOP and SHOWTYPE. 

Getting an error on the following item..
```
00057  000000                                      (1/40)55          SHOWOP SAVE                   * S
00058  000000                                      (1/19)56+         MNOTE 'O'SAVE = U'
```
I am reading that the 2nd line should read `MNOTE 'O'SAVE = S` but is showing U.
Is that correct

abekornelis: That's strange! On my test I get the following lines on PRN:
```
55          SHOWOP SAVE                   * S
56+         MNOTE 'O'SAVE = S'
```
and later on - when SAVE is renamed to SAVREGS:
```
 98 SAVREGS  OPSYN  SAVE
 99 SAVE     OPSYN
100          SHOWTYPE SAVREGS              * S
101+         MNOTE 'O'SAVREGS = S'
103          SHOWTYPE SAVE                 * U
104+         MNOTE 'O'SAVE = U'
```
Both are entirely correct as far as I can see. 
I did update the source with additional comments. Please pull to retrieve latest version. Then please retry.

adelosa: OK.. just pulled down your last 2 commits, ran bldjar and bldlib and still getting the same result.
I have attached full output for your reference. Obviously something different between yours and mine.
[output.txt](https://github.com/z390development/z390/files/11569743/output.txt)
Happy to assist to try and work out what is going on.

adelosa: What are your asm params you are using and could that make a difference?

adelosa: BTW, I pushed the additional tests and it also fails on GitHub... so you must be running things differently than me and GitHub.. Let me know if you need help working this out.

abekornelis: from within the groovy environment the assembler seems to be unable to find the SAVE.MAC file, which is located in the mac subdirectory of z390. The problem reproduces when I run the test from Groovy.
When I run the test using ASM.BAT the invocation adds default path for macros and copy members on the call to mz390:
` sysmac(+%z_HomeDir%mac) syscpy(+%z_HomeDir%mac)`

Could it be (I have no clue how to check this) that the call to z390 from Groovy does not add these two parameters?

adelosa: yes.. that will be it.. did not notice an error for a missing macro.. Will add the sysmac which will fix the test.

adelosa: The test framework does not exactly mirror the bat files. You need to provide any additional parameters like sysmac. It could be changed to mimic, but my personal preference is to provide it explicitly.

abekornelis: Anthony, there is no error for a missing macro - you did not miss it.
There would have been an error if we'd tried to expand the macro.
Yet in this case we're only asking for the O' attribute of a label 'SAVE' - z390 looks for opcodes, macros, etc. finds nothing and therefore evaluates to U for Unkown/Undefined.
Quiet different from coding a call to the macro which must fail when it is not found.
If you add the missing parm for SYSMAC we should be okay. SYSCPY should not be needed.

adelosa (dismissed):
I'm happy with the testing if you are. I can't really comment on the essence of the change itself - that would best come from @jyganci or @dsh33782. My only other comment is do these changes require any updates to the docs?

abekornelis:
No doc changes needed as this fixes a bug in z390. I'll wait with the merge until after the weekend to give @jyganci and @dsh33782 an opportunity to comment, if they wish to. Otherwise I'll try to merge the Pull Request after the weekend.

src/mz390.java
```diff
@@ -11638,27 +11639,46 @@ private void exec_pc_pfx_o(){
...
-					if (opcode_type == 3
-							|| opcode_type == 6
-							|| opcode_type == 13){
-						setc_value = "E"; // BRX, BCX, BRCX extended mnemonic
-					} else {
-						setc_value = "O"; // machine opcode
-					}
+                    index = tz390.find_key_index('O',key);                                          // #485
+                    //System.out.println("afk mz390.exec_pc_pfx_o OPCODE index=" + index);          // #485 **!!
+                    if (index > -1)                                                                 // #485
```
jyganci: Suggestion only. Nothing wrong. For consistency with previous checks after tz390.find_key_index() invocations, replace "index > -1" with "index >= 0".

z390test/src/test/groovy/org/z390/test/TestOAttrib.groovy
```diff
@@ -0,0 +1,64 @@
...
+class TestOAttrib extends z390Test {
+
+    var options = ["SYSMAC(${basePath("mac")})"]
+
+    @Test
+    void test_OAttrib() {
+        this.printOutput = true
+        int rc = this.asm(basePath('rt', 'mlc', 'ATTRIB$O'), *options)
+        this.printOutput()
+        assert rc == 0
```
jyganci, line 16: Observation. If the sent sample TOP.MLC is used as a model to update ATTRIB$0.MLC, then this test could be simplified to be

```
void test_OAttrib() {
    int rc = this.asmlg(basePath('rt', 'mlc', 'ATTRIB$0'), *options)
    this.printOutput()
    assert rc == 0
}
```

src/tz390.java
```diff
@@ -1216,6 +1224,20 @@ public  class  tz390 {
 //           Alternate syntax is ;*Short which will generate only H, L, E, NH, NL, NE rather than the full set
 //           This syntax allows us to override the default mnemonics for generated instructions.
 //           Omission of mnemon in the override implies that no  mnemonic exists for that mask.
+// Example1: definition of assembler directive AIF: "--=AIF,203,--"                                                                                                    #485
```
jyganci, line 1305: Observation/question. The example has -- for opcode and op_trace_type. Nothing wrong. However, Remark 1 above (not part of your changes) says "op_type and op_trace_type are entered as -- for assembler directives". Should this be changed to "opcode and op_trace_type are entered as -- for assembler directives"???

src/tz390.java
```diff
@@ -1216,6 +1224,20 @@ public  class  tz390 {
...
+// Example2: definition of instruction CLC:         "D5=CLC,17,170"     --> opcode D5,  mnemonic CLC,  opcode type 17 (see table opcode_formats above), trace type 170 #485
+// Example3: definition of instruction BC:          "47=BC,5,50"        --> opcode 47,  mnemonic BC,   opcode type  5 (see table opcode_formats above), trace type  50 #485
+// Example4: definition of B, BE, BNZ, etc:         "47m=Bm,6,60;0=NOP" --> opcode 47m, ext.mnemonics, opcode type  6 (see table opcode_formats above), trace type  60 #485
+//                                                                          generated extended mnemonics are Bx, BNx except for mask=0 mnemonic NOP is forced          #485
+// Example5: definition of instruction JLC:         "C04=JLC,33,330"    --> opcode C04, mnemonnic JLC, opcode type 33 (see table opcode_formats above), trace type 330 #485
```
jyganci: This example should be deleted or updated to match your new line 2077 below. If  you update it (recommended), note the misspelling "menmonnic".

z390test/src/test/groovy/org/z390/test/TestOAttrib.groovy
adelosa: @jyganci this is an interesting comment and poses a question for the project - where to write tests? The test framework is designed to perform testing easily but does not stop you writing ASM to test as you have done with TOP.
Maybe worth a discussion at next meeting.

jyganci:
Anthony,

Not sure what you mean. A lot of my tests work this way. For example, look
at RunAssist.groovy, built from bash/runassist. The tests test_TSTXDECI()
and test_TSTHEXI()  invoke asmlg for TSTXDECI.MLC, respectively
TSTXHEXI.MLC. The programs are similar to TOP.MLC.

Are we not still in "dual mode", writing bash/bat tests, then incorporating
them into the test framework?

What am I missing?

jyganci:
Abe, I will approve this after I hear that you've looked over my comments. Mostly suggestions, observations or typos. Nothing wrong. The suggested change to the test program is just something to think about. What you and Anthony did is for the groovy test is fine. Do you plan to merge this as is and work on the vector instructions in a separate pull request?

src/tz390.java
abekornelis: The format of entries is documented as:
> opcode=mnemonic,op_type,op_trace_type[;overrides]

the table entry given in example 1 is "--=AIF,203,--"
So AIF is the mnemonic, 203 is the op_type. Both opcode and op_trace_type are specified as dashes --

In my reading this is consistent with the current wording of remark 1. 
Yet I am open to a different way of wording the information if that reduces the risk of the comments being misunderstood or misinterpreted. The aim of these comments is to avoid confusion, rather than create it ...

src/tz390.java
abekornelis: I have updated the example to match the actual source code and modified a few details in the comments.

src/tz390.java
jyganci: "Example1: definition of assembler directive AIF: "--=AIF,203,--""
opcode=--
mnemonic=AIF
op_type=203
op_trace_type=--

"Remark 1: op_type and op_trace_type are entered as -- for assembler directives"

How are these consistent?

src/tz390.java
jyganci: "// Example5: definition of instruction JLC:         "C04=JLC,33,330"    --> opcode C04, mnemonnic JLC, opcode type 33 (see table opcode_formats above), trace type 330 #485"

tz390.java line 2077:  "C04=JLC,33,330;*Extended", //   "C04"  "JLC"     "BLX"  33 RPI 2221 #485

src/tz390.java
jyganci: Please address the preceding comment.

jyganci (approved):
While the tz390.java "Remark 1 and Example 1" inconsistency remains, it is not part of your changes, so approval is granted.

Issue #476, opened 2023-03-04
Bat file vsam1/demo/bat/KSF1RPO2.BAT abnormally terminates trying to unload PATH NAMELIST


jyganci:
To reproduce, uncomment the "GOTO skipnl" statement on KSF1RPO2.BAT line 80  (Windows) or uncomment the unload NAMELIST code in vsam1/demo/bash/ksf1rpo2 and run the bat/bash script. Sample output when run on Linux:

```
before unload NAMELIST
14:07:30 REPRO     EZ390 START USING z390 v1.8.1 ON J2SE 17.0.5 02/28/23
REPRO z390 VSAM UTILITY V1.2
REPRO UNLOADING VSAM INFILE TO QSAM OUTFILE
EZ390E error  43 I/O error for DCB=000FE050 DDNAME=VXN#2 FILE=
EZ390E error  23 i/o error on open - java.io.FileNotFoundException:  (No such file or directory)
EZ390I Trace Table Entry      800FE866 92D8C741     MVI  
EZ390I Trace Table Entry      800FE86A 12FF         LTR  
EZ390I Trace Table Entry      800FE86C 4770D2B0     BNE  
EZ390I Trace Table Entry      800FE880 D201D698D2BA MVC  
EZ390I Trace Table Entry      800FE886 47F0D69A     B    
EZ390I Trace Table Entry      800FEC6A A715001A     BRAS 
EZ390I Trace Table Entry      800FEC9E 0A23         SVC  
EZ390I Trace Table Entry      800FECA0 4110C538     LA   
EZ390I Trace Table Entry      800FECA4 41000040     LA   
EZ390E error  11 ABEND PSW=07852600 800FECA8 0A13         SVC   ABEND S013
 R0-R3 00000000000001A4 00000000000FDFE8 00000000000FE0AD 0000000000000000
 R4-R7 0000000000000000 0000000000000000 0000000000000000 0000000000000000
 R8-RB 00000000000FE0B8 00000000000FE0B8 0000000000000000 0000000000000000
 RC-RF 00000000000FF5D0 00000000800FE5D0 0000000000002018 0000000000000000
EZ390E error  12 program aborting due to abend S013
14:07:30 REPRO     EZ390 ENDED   RC=16 SEC= 0 MEM(MB)= 17 IO=203 INS=122
exec ERROR: ez390 rc=16; see errors in linklib/REPRO.LOG file and on console
ERROR: error in unload of NAMELIST; exiting
```
[...]

PR #489, opened 2023-05-19
vz390.java fix and updated bash, bat files for issue #476

jyganci:


abekornelis:
I have looked at it - seems okay from my perspective. But I do not very well understand the old zVSAM code. I'll leave it to Don to either approve or suggest additional changes.

abekornelis (approved):

Issue #503, opened 2023-07-02
Define new z16 instructions in opcode tables


abekornelis:
This issue is created to break up issue #414 into more manageable chunks

In tz390 implement support for MACHINE and OPTABLE options for the new z16 machine.
Create new opcode table to define all new mnemonics.

WARNING:
Do not yet update opcode table UNI to include z16 instructions. 
This is to be done after all assembly and emulation support has been implemented.

Issue #486, opened 2023-05-11
some supported instructions defined in opcode definition tables for unsupported instructions


abekornelis:
The opcode definition tables documenting unsupported instructions contain some entries for instructions that are in fact supported.
Let's clear out these unneeded and confusing entries.

PR #511, opened 2023-07-05
Changes to opcode tables for z16 support

abekornelis:
solves #503 
solves #486


doc/getting_started/gradle_commands.txt.txt
adelosa, line 1: I think this type of comment lives in the docs/contributing/contributing_code.md file rather than being a new file.

jyganci: File name ends with "txt.txt" ???

src/tz390.java
```diff
@@ -3348,11 +3361,85 @@ public void init_tz390(){
...
+            // if the optable is an alias, issue error when the alias is not defined                     // #503
+            if (!optable_option_id[index].equals(optables_optable[index]))                               // #503
+               {index2=0;                                                                                // #503
+                while (index2 < index) // check all entries defined before this one                      // #503
+                   {if (optables_optable[index].equals(optables_optable[index2]))                        // #503
+                       {break;                                                                           // #503
+                        }                                                                                // #503
+                    index2++;                                                                            // #503
+                    }                                                                                    // #503
+                if (index2 >= index) // base optable is defined?                                         // #503
```
jyganci: "base optable is defined?" should be ??? "base optable is not defined?"

src/tz390.java
```diff
@@ -3348,11 +3361,85 @@ public void init_tz390(){
...
+            machines_optable [index]=entry.substring(i+1);                                               // #503
+            // issue error when the optable is not defined                                               // #503
+            index2=0;                                                                                    // #503
+            while (index2 < optables_optable.length) // check all optable entries defined                // #503
+               {if (machines_optable[index].equals(optables_optable[index2]))                            // #503
+                   {break;                                                                               // #503
+                    }                                                                                    // #503
+                index2++;                                                                                // #503
+                }                                                                                        // #503
+            if (index2 >= optables_optable.length) // optable is defined?                                // #503
```
jyganci: "optable is defined?" should be ??? "optable is not defined?"

src/tz390.java
abekornelis: I have changed the comment to: invalid index indicates not-found condition

src/tz390.java
abekornelis: I have changed the comment to: invalid index indicates not-found condition

jyganci: Comment not changed??? My interpretation of the "//comment" on the "if ..." line

if ( condition )  // comment
{
    code if condition is true
}

explains what condition==true means. In your case, "index2 >= optables_optable.length" is checking to see if the optable is not defined -- no match in the while loop. 

I guess we disagtee on what the comment should say.

jyganci (dismissed):

src/tz390.java
abekornelis: Changed comments have not yet been pushed. Waiting for Anthony to complete our discussion on the git and gradle commands quickref. I'll push the changes once all is done. Thanks for your approval.

doc/contributing/contribute_code.md
adelosa, line 1: I can live with this as-is but if it was me, this is what I would change... 
1. I would not have git commands - there are enough on the net about Git.
2. I would move the gradle details to the top after the `rebuild the jar` section and call it `test the jar`
Its your call..

abekornelis: Anthony, I get what you're at. But I need a place where I can quickly find a tailored quickref for z390-related git and gradle commands. That's why I originally put them in separate txt files. I'll move the gradle section up, though. Once done, I'll wrap it up and push the final update to the repo for PR completion.

jyganci (approved):

PR #519, opened 2023-08-20
Automated GitHub build restructure

adelosa:
The following PR is to replace the existing automated build for the project.

The issues with the current build:
* Tests are completely run twice - for Linux and Windows. Given it is Java (write once, run anywhere) a full regression is not necessary for each platform.
* Jar is recreated multiple times. Binary included in distribution is not actually tested
* Distribution has work files that are generated as part of the test process.
* Current PDF formatting is not fantastic - removing its inclusion until I find something better

New build takes the following approach:

Workflow: build.yaml
* Job 1:  Build z390 distribution zip file by using bash/blddist script - save the Distribution zip
* Job 2: Run full regression tests on Linux using distribution zip from job 1
* Job 3: Run ASM and Cobol demo scripts on Windows using distribution zip from job 1
* Job 4: If Tag - create a GitHub release using the distribution zip from job 1

Workflow: docs.yaml
* Job 1: Build HTML docs from project doc folder and publish to GitHub pages (no changes except updates to py version)


adelosa:
*** Need to fix tests before merge ***

The gradle tests are run against the jar and source in the distribution zip.
So build gets zip, extracts the content and then runs the tests against that source.
This causes the folder name to the source to be longer which causes the source to not assemble/compile
because the line is too long.

There are 3 such tests in the testbase.
* RunCblDemos.groovy -> zCOBOL/demo/COPYFILE
* RunCmdProc.groovy -> rt/test/TESTCMD3
* RunCmdProc.groovy -> rt/test/TESTCMD4

These tests need to be updated to remove the filenames and use files referenced in envvars instead.

------

Make sure to squash before merging as there are lots of work commits in this pull request as testing involves pushing to GitHub to run the actions.

I fixed an issue with the javadoc build `bldjavadoc`- it was placing its output in the `docenv` folder which should only be used for the python virtual environment that the doc build uses - I have changed it to use `build/javadoc` folder instead.
I also added a param for `bldjavadoc` to provide an optional output folder in case you don't want to use `build/javadoc`. 
I also removed the execution of `bldjavadoc` from `bldjar` and I now execute it from `blddist` instead. 

If you use the javadocs, then you will need to rerun bldjavadoc after bldjar. I have not checked the BAT jobs for consistency.. @abekornelis - your thoughts would be appreciated as I think you added the javadoc build into the bldjar function. For the build, it makes things easier not having multiple levels of scripts.

abekornelis:
Anthony, I did indeed add the javadoc gen step to the jar build. John was not too happy with the details of the implementation, if I remember correctly. In hindsight I've had my own misgivings. I'm entirely fine if you take it out again. The javadoc can still be built whenever needed. As part of the build process would be required. Whether or not for a distro. A build should build the full product including all bells and whistles. Fine with me if you change it. Or if you create an issue to make the change at some later time.

adelosa:
I have tested this now with @jyganci changes to remove the hard coded filenames. Build is now working. 
Ready for review and merge if OK.

bash/blddist
```diff
@@ -7,71 +7,68 @@ set -e                                  # exit on all errors
 cd $(dirname $0)                        # set to directory of script
 cd ..                                   # cd back to root       
 
-root_dir=$(pwd)
-version_tag=$(git describe)
-dist_folder=dist
+ROOT_DIR=$(pwd)
+VERSION="$(git describe --tags)"
+VERSION="${VERSION//./_}"
```
jyganci: Question only. Is this line here (also line 18 of build.yml) because of issue 518 zc390.java fails when directory name contains a period? If yes, a potential fix exists. See issue 518 comment added Oct 7.

bash/blddist
```diff
@@ -7,71 +7,68 @@ set -e                                  # exit on all errors
...
 # add core files to the build dir
-cp ${root_dir}/z390.jar .               # jar
-cp ${root_dir}/site/pdf/z390.pdf .      # pdf doc
-core_dirs=( bash bat mac zcobol cics linklib README.txt LICENSE )
-for dir in ${core_dirs[@]}; do
-    cp -R ${root_dir}/${dir} .          # unix bash scripts
+cp ${ROOT_DIR}/z390.jar .               # jar
+CORE_DIRS=( bash bat mac zcobol cics linklib README.txt LICENSE )
+for dir in ${CORE_DIRS[@]}; do
+    cp -R ${ROOT_DIR}/${dir} .          # unix bash scripts
```
jyganci: Remove the "# unix bash scripts" comment.

bash/blddist
```diff
@@ -7,71 +7,68 @@ set -e                                  # exit on all errors
...
 done
 
 # add test and supplimental components
-supp_dirs=( assist demo guam qsam bsam rt soa sort structuredmacros tests vsam1 
+SUPP_DIRS=( assist demo guam qsam bsam rt soa sort structuredmacros tests vsam1 
 vsam2 vse zopcheck zpar )
-for dir in ${supp_dirs[@]}; do
-    cp -R ${root_dir}/${dir} .          # unix bash scripts
+for dir in ${SUPP_DIRS[@]}; do
+    cp -R ${ROOT_DIR}/${dir} .          # unix bash scripts
```
jyganci: Remove the "# unix bash scripts" comment.

bash/blddist
```diff
@@ -7,71 +7,68 @@ set -e                                  # exit on all errors
...
+# add javadoc
+mkdir -p javadoc
+cp -R ${ROOT_DIR}/build/javadoc/* javadoc
 
 # remove components not required for distribution
-del_items=('bash/bldjar' 'bat/BLDJAR.BAT' 'bash/blddist' 'bash/relver' 'bat/RELVER.BAT')
-for item in ${del_items[@]}; do
+DEL_ITEMS=('bash/bldjar' 'bat/BLDJAR.BAT' 'bash/blddist' 'bash/relver' 'bat/RELVER.BAT')
+for item in ${DEL_ITEMS[@]}; do
     rm ${item}        # unix bash scripts
```
jyganci: Remove the "# unix bash scripts" comment.

jyganci (commented):
I'm ready to approve after you've responded to comments added to the PR.

bash/blddist
adelosa: I have added your fix and removed the workaround.

bash/blddist
adelosa: Removed the comment

bash/blddist
adelosa: Removed

bash/blddist
adelosa: Removed

adelosa:
Hi @jyganci - I have updated according to your comments and incorporated the fix for the dot in directory name.

abekornelis:
I am fine with these changes. I will let John approve the PR as you both have spent considerable more time and effort than I did.

.github/workflows/build.yml
```diff
@@ -2,37 +2,95 @@ name: Build z390
...
+        uses: actions/download-artifact@v3
+        with:
+          name: dist
+      - name: Install dist and run tests
+        run: |
+          mkdir -p dist
+          unzip z390_${VERSION}.zip -d dist
+          ls -l dist/z390_${VERSION}
+          cp dist/z390_${VERSION}/z390.jar .
+          Z390_PROJECT_ROOT="../dist/z390_${VERSION}" z390test/gradlew -p z390test test
```
jyganci, line 52: Question only. This looks like 2 statements: set Z390_PROJECT_ROOT and run z390test/gradlew. You can have 2 statements on the same line? I'm probably missing something.

jyganci (approved):
Looks good. I did ask one question. Answer that and merge the PR.

.github/workflows/build.yml
adelosa: 2 statements. One sets the envvar then runs the command. I believe the envvar scope is just for that command. I see it as an extension of the command.

jyganci:
Interesting. Learned something new.

Issue #520, opened 2023-10-06
Remove hard-coded file names in source modules used in regression tests


jyganci:
COBOL source demo/COPYFILE.CBL, rt/test/TESTCMD3,MLC and rt/test/TESTCMD4.MLC contain hard-coded names for external files used for testing. This causes issues in the Groovy regression test scripts that use thise programs. Changes required:
1. Change COPYFILE.CBL to use DDNAMEs INFILE and OUTFILE instead of the hard-coded names.
2. Change bash/runcbldemos and bat/RUNCBLDEMOS.BAT to set environment varibles INFILE, OUTFILE containing these names.
3. Change TESTCMD3.MLC and TESTCMD4.MLC to no longer determine whether Windows or Linux/MacOS to determine test file names. Instead, get file names from environment variables.
4. Change bash/runcmdproc and bat/RUNCMDPROC.BAT to set the appropriate environment variables.
5. Modify regression test files RunCblDemos.groovy and RunCmdProc.groovy to mimic the bash,bat changes described in 2 and 4.

adelosa:
Thanks for doing this work John. The changes you are proposing look fine to me. 
I have reviewed your groovy test scripts and they are perfect... and it removes the complexity of dynamically modifying the scripts that was there before. 

Can you create a pull request with these changes and then I can run them locally and review the actual code here on GitHub?

I'll have a look at the Gradle Deprecation warning message and suggest a fix once you get your pull request completed.
https://docs.gradle.org/current/userguide/upgrading_version_8.html#test_framework_implementation_dependencies
Looks like it can be fixed easily by adding a dependency to the JUnit test library.

I'll rebase my #519 PR on top of this change once it's merged to main.
Thanks again @jyganci !

jyganci:
All,

Pull request #521 that fixes issue #520 created. This is for the problems
reported by Anthony with regard to some Groovy scripts in the regression
test set.

Note: Even though we are probably 99% confident of the new Groovy
regression test setup, I'm still adding the old style bash and bat file
tests along with the corresponding modified Groovy scripts.

Anthony: I assigned you as reviewer.

Regards, John Ganci

jyganci:
Fixed by PR #521

PR #521, opened 2023-10-07
issue #520 delete hard-coded file names in bash,bat,src,groovy for regression tests

jyganci:


bash/runcbldemos
```diff
@@ -5,23 +5,27 @@ set -e              # exit on all errors
 cd $(dirname $0)    # set to directory of script
 cd ..
 
-bash/cblclg zcobol/demo/HELLO $1 $2 $3 $4 $5 $6 $7 $8 $9
+demopath=zcobol/demo
```
adelosa, line 8: No issue, just curious. Why did you add the demopath variable for the new version?

bash/runcmdproc
```diff
@@ -23,17 +23,37 @@
 # debug flag; 0=no debug, 1=debug
 debug=0
 
-set -e              # exit on all errors
+#set -e              # exit on all errors
```
adelosa: Do you really want to never stop on errors.. I would prefer if you `set +e` when you want to handle return codes and then `set -e` again when you are finished.

adelosa (dismissed):
No issues, just a couple of queries.. I think this has actually simplified the tests and reduced the amount of code which is nice.

jyganci:
1. Usage of demopath (and rtpath). Melvyn used cicspath variables (and
others) in the CICS bat files. It minimizes changes if the value it
represents is used in several places. I've started using the "technique" in
general. Plus, it kind of mimics the "basePath('abc",'def',...)" usage in
the Groovy scripts.

2. Commenting the "set -e" in bash/runcmdproc. My reasoning was so that the
script would have a chance to issue an error/warning message if an invoked
script failed. Rather than sprinkle "set +e", "set -e", put it back the way
it was. Change has been pushed to my fork.

Regards, John Ganci

adelosa (approved):
Ready to merge!

adelosa:
Merged to main

PR #528, opened 2024-06-05
is525 Register 14 not set for EODAD or SYNAD

jyganci:
IBM documentation states that upon entry to either an EODAD or SYNAD exit for QSAM or BSAM, register 14 contains the address of the instruction following a GET (QSAM) or CHECK (BSAM). The z390 code in sz390.java did not set register 14 when processing a GET or CHECK that resulted in EODAD or SYNAD being invoked. Code has been added to sz390 to do so. In addition, a new test program, demo/TESTDCB2.MLC (a modified version of demo/TESTDCB1.MLC), has been added to the test suite that has an EODAD return point that uses the register 14 value to return to the GET that resulted in end of input.

### DIFF
--- a/demo/TESTDCB2.MLC
+++ b/demo/TESTDCB2.MLC
@@ -1,6 +1,6 @@
 *********************************************************************
 * z390 - Mainframe assembler emulator and run-time engine
-* Copyright (C) 2021 z390 Assembler LLC
+* Copyright (C) 2024 z390 Assembler LLC
 *
 * This file is part of z390.
 *
@@ -16,7 +16,7 @@
 * You should have received a copy of the GNU General Public License 
 * along with this program; if not, see https://www.gnu.org/licenses.
 
-* Author - Don Higgins                                              *
+* Author - Don Higgins and John Ganci                               *
 * Date   - 06/04/2024                                               *
 *********************************************************************
 * Same as demo/TESTDCB1 but illustrates EODAD that returns to GET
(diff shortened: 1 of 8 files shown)


demo/TESTDCB2.MLC
```diff
@@ -0,0 +1,64 @@
...
+* (at your option) any later version.
+* z390 is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License 
+* along with this program; if not, see https://www.gnu.org/licenses.
+
+* Author - Don Higgins                                              *
```
abekornelis: you might want to change the author (and the copright date at the top, too)

demo/TESTDCB2.MLC
```diff
@@ -0,0 +1,64 @@
...
+ELOOP    DS    0H
+         CLOSE MF=(E,CFILES)
+         WTO   'TESTDCB2 Ended OK'
+         SUBEXIT
+***********************************************************************
+*        EODAD routine
+***********************************************************************
+EOF      DS    0H
+         MVI   EOFFLAG,C'Y'        Set end of file reached on input
+         BR    14                  Return to GET
```
abekornelis: Since the plan is to test that R14 points to after the GET instruction, consider a CL with an Acon for the expected address. Having two GET instructions makes it more complicated. Having only 1 GET (at the top of the loop) might make this a bit easier...

abekornelis (changes requested):
I've added a few comments. Please see those for details.

abekornelis (approved):

PR #536, opened 2024-06-13
Optable 370

abekornelis:
Fixing optable(370) to ensure compatibility with HLASM.
Mnemonic ACALLPRM had to be moved to optable for z390 extensions.
As a result the ACALLPRM is no longer available under default optable(UNI).
Therefore any program using ACALLPRM (mainly zCobol programs) need to have optable(z390) specified explicitly.
Resulting in many fixes for scripts (bat, bash, groovy)


cics/CICSCBL.OPT
```diff
@@ -0,0 +1,6 @@
+  cics
+  optable(Z390)
+  stats
+  syscpy(cics+zcobol/CPY+zcobol/MAC+MAC+.)
+  sysmac(cics+zcobol/MAC+MAC)
```
jyganci, line 5: Change MAC to mac. Four occurrences. Left as is will break bash scripts.

This appears many times in these OPT files. I will leave it up to you to find and change them.

cics/Z390RMTE.OPT
```diff
@@ -0,0 +1,4 @@
+  ascii
+@cics/CICSDFLT.OPT
+  syscpy(MAC)
+  sysmac(MAC)
```
jyganci, line 4: MAC should be mac.

cics/Z390SEQ.OPT
```diff
@@ -0,0 +1,4 @@
+  ascii
+@cics/CICSDFLT.OPT
+  syscpy(MAC+.)
```
jyganci, line 3: MAC -> mac.

cics/CICSDFLT.OPT
```diff
@@ -0,0 +1,5 @@
+  optable(Z390)
+  stats
+  syscpy(cics+MAC)
+  sysmac(cics+MAC)
```
jyganci, line 4: MAC -> mac.

adelosa:
Hi Abe,
I think from a procedural perspective, the fixing of the optables is a separate request from changing inline options to use option files. I wish these were 2 separate pull requests - a case of a PR getting too big 😀

Comments below relate to the change to use option files everywhere.

I am not a fan of option files:
1. They hide details of the options used in a command. Use of nested option files makes this even worse. This point was made by @jyganci and I agree.
2. Some options include paths which limits their re-usability unless you can set the root which is not currently possible.

If users of z390 want to use options, they can. I'm not against users having access to options files.
But for the library/app, I would like any path specific options to be removed. Such options should be specified in the context of the bash/bat script that is executing the code as it knows what the root is and can set it accordingly.

If you did want to use option files, then we would need a way of setting the root to be used.. 
`bash/asm zcobol/lib/ABORT OPTROOT(some/path) @$optfile $1 $2 $3 $4 $5 $6 $7 $8 $9`

To me, thats still not great either.. and probably a lot of work for something many people don't want or need.

In summary, I don't support this PR because of the option changes. I know you mentioned this in our last call but I was not expecting such a BIG change.. if it was 1-2 option files, I'd be OK but this is a wholesale change in approach to how we provide options to commands in the app and needs some more discussions to get across the line.

abekornelis:
This pull request is replaced by PR#537

PR #553, opened 2024-08-13
Issue#545 Extend generated java doco to include private methods

abekornelis:


bat/BLDJVDOC.BAT
```diff
@@ -15,7 +15,7 @@ pushd %~dps0..
 rem generate javadoc
 if not exist ".\docenv\javadoc\" mkdir docenv\javadoc
 cd docenv\javadoc
-call javadoc.exe ..\..\src\*.java || goto error
+call javadoc.exe ..\..\src\*.java -private || goto error
```
jyganci, line 18: Add similar change to bash/bldjavadoc. Line 20 there. New line 20 could be
```
javadoc -quiet -Xdoclint:none -d ${DEST} -private src/*.java
```

src/ez390.java
```diff
@@ -294,7 +295,7 @@ public void process_ez390(String[] args,JTextArea log_text,JTextField command_te
     /**
      * execute IPL pgm and/or application pgm
      *
-     * @param zcvt_pgm_addr
+     * @param zcvt_pgm_addr - ??
```
jyganci: Why did you add " - ??"?

bat/BLDJVDOC.BAT
abekornelis: Changed as per your suggestion. Thanks.

src/ez390.java
abekornelis: The javadoc utility complained about missing description. I had no clue. So I put in question marks.
Admittedly a tad cryptic. I'll add proper text instead.

src/ez390.java
```diff
@@ -294,7 +295,7 @@ public void process_ez390(String[] args,JTextArea log_text,JTextField command_te
     /**
      * execute IPL pgm and/or application pgm
      *
-     * @param zcvt_pgm_addr
+     * @param zcvt_pgm_addr - function of this parameter is currently unclear
```
jyganci: After looking in ez390 to see where run_pgm is invoked, ez390 line 297 should be something like
```
     * @param zcvt_pgm_addr  address of 8 byte field containing name of program to run
 ```

jyganci (approved):

Issue #554, opened 2024-08-14
Correct OPTABLE(ESA,LIST) output to match HLASM


abekornelis:
The list of supported mnemonics when running with OPTABLE(ESA) is incorrect. It should match HLASM's output for the option.

abekornelis:
Fixed by PR #555

PR #555, opened 2024-08-21
Issue 554 Fix optable differences for optable(ESA)

abekornelis:
Fixes Issue #554


rt/bat/RUNOPTABLES.BAT
```diff
@@ -14,88 +14,283 @@ rem ----- Lvl(%z_NestLevel%) Start %0 %1 %2 %3 %4 %5 %6 %7 %8 %9
...
+                         set /A count_ok=%count_ok%+1
+                         ) else (
+                         echo %0 ERROR: OPTB# returned %z_ReturnCode% for "machine(ARCH-4)"
+                         set /A count_err=%count_err%+1
+                         if (%z_ReturnCode% GTR %z_MaxRetCode% (set z_MaxRetCode=%z_ReturnCode%)
+                         )
+if %1. NEQ . goto return
+
+:report
+echo %count_ok% tests were okay and %count_err% failed
```
jyganci, line 293: :Lot of work here!

What follows is just a suggestion. Not required. ONLY DO THIS IF YOU WANT TO DO IT!

You could have 2 (count-ok, count-err) pairs -- one for optable, one for machine. At label :report echo the optable subtotals and echo the machine subtotals, followed by and echo of the grand totals.

src/az390.java
```diff
@@ -1878,17 +1882,28 @@ else if (tz390.op_trace_type[index]==102)
...
+                    if (my_hexop.length() == 3) // mask excluded?                      // #554
+                       {my_hexop=my_hexop.substring(0,2)+"."+my_hexop.substring(2,3);  // #554
+                        short_op = 1;           // mark short opcode                   // #554
+                        }                                                              // #554
+                    else                        // mask included!                      // #554
+                       {my_hexop=my_hexop.substring(0,2)+my_hexop.substring(3)+my_hexop.substring(2,3); // #554
+                        short_op = 0;           // mark full opcode                    // #554
+                        }                                                              // #554
+                    if (tz390.op_trace_type[index]==130)                               // #554
+                       {if (short_op == 1)      // short opcode                        // #554
```
jyganci, line 1899: Just a suggestion. Nothing wrong.

You could define short_op as a boolean and set to true or false instead of 1 or 0, replacing `if *short_op == 1)` with `if (short_op)`. If you want to be slick, replace the `if ... else ...` with one line
```
my_operands = (short_op ? "R1,R12" : "R12");
```
Or, just leave it as is. If you decide to change, there are a few more short_op usages below.

src/tz390.java
```diff
@@ -2173,6 +2172,12 @@ public  class  tz390 {
          "A7D=MGHI,73,731",      //   2470 "A7D"   "MGHI"     "RI"   12 // RPI 1522
          "A7F=CGHI,73,731",      //   2490 "A7F"   "CGHI"     "RI"   12 // RPI 1522
          "B250=CSP,14,140",      //   3150 "B250"  "CSP"      "RRE"  14
+         "B280=LPP,7,70",   // S,LPP,D1(B1)   RPI 2221
+         "B284=LCCTL,7,70", // S,LCCTL,D1(B1) RPI 2221
+         "B285=LPCTL,7,70", // S,LPCTL,D1(B1) RPI 2221
+         "B286=QSI,7,70",   // S,QSI,D1(B1)   RPI 2221
+         "B287=LSCTL,7,70", // S,LSCTL,D1(B1) RPI 2221
+         "B28E=QCTRI,7,70", // S,QCTRI,D1(B1) RPI 2221
```
jyganci, line 2180: Question only. You decide. Since you deleted these lines above and moved them here, do you need to add #554 in the commnet?

jyganci (approved):
There are some comments/suggestions/questions. Please read them over and decide if you want to make any of the changes. If not, go ahead and merge.

src/az390.java
abekornelis: John, I never thought of the ternary operator. I find it confusing. I guess that merely shows my lack of experience with Java and C/C++
I prefer to leave as is, although I agree that it looks better from a coding perspective.

PR #570, opened 2024-09-11
Changes for optable YOP

### LINKED ISSUE #568: Correct OPTABLE(YOP,LIST) output to match HLASM
The list of supported mnemonics when running with OPTABLE(YOP) is incorrect. It should match HLASM's output for the option.

abekornelis:
Resolves #568 Correct OPTABLE(YOP,LIST) output to match HLASM

### DIFF
--- a/src/az390.java
+++ b/src/az390.java
@@ -2433,11 +2433,11 @@ public class ReportEntryComparator implements Comparator<String>                
    {// special case: entries consisting of a single space sort first!              // #568
     if (str1.charAt(0) == ' ') return 1;  // blank entry always smaller            // #568
     if (str2.charAt(0) == ' ') return -1; // blank entry always smaller            // #568
-    for (int i=0; i<7; i++)                                                        // #568
+    for (int i=0; i<8; i++)                                                        // #568
      {char c1 = str1.charAt(i);                                                    // #568
       char c2 = str2.charAt(i);                                                    // #568
-      if (c1 == c2) {} // continue on to next char                                 // #568
-      else // we found different characters                                        // #568
+      // if c1 == c2 then we continue on to the next character                     // #568
+      if (c1 != c2)                                                                // #568
        {if (c1 == ' ') return -1; // Mnemonic 2 is longer                          // #568
         if (c2 == ' ') return 1;  // Mnemonic 1 is longer                          // #568
         if (c1 <= '9')                                                             // #568
--- a/src/az390.java
+++ b/src/az390.java
@@ -31,6 +31,7 @@ import java.math.RoundingMode;
 import java.nio.ByteBuffer;
 import java.text.SimpleDateFormat;
 import java.util.Arrays;
+import java.util.Comparator;                                // #568
 import java.util.Date;
 import java.util.Iterator;
 import java.util.TreeSet;
@@ -427,6 +428,7 @@ public  class  az390 implements Runnable {
         * 2024-08-12 #545 Extend generated java doco to include private methods
         * 2024-08-15 AFK #554 Correct OPTABLE(ESA,LIST) output to match HLASM
         * 2024-08-23 AFK #561 Correct OPTABLE(ZOP,LIST) output to match HLASM
+        * 2024-09-09 AFK #568 Correct OPTABLE(YOP,LIST) output to match HLASM
 	*****************************************************
     * Global variables                        last rpi
     *****************************************************/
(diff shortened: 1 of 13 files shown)


src/az390.java
```diff
@@ -2410,6 +2421,41 @@ else if (tz390.op_trace_type[index]==613)                          // #533
...
+ * The mnemonic portion occupies the first 8 positions of the ReportEntry
+ * and contains only uppercase characters and digits.
+ * Digits should sort higher than alphabetic characters.
+ */
+public class ReportEntryComparator implements Comparator<String>                   // #568
+ {@Override                                                                        // #568
+  public int compare(String str1, String str2)                                     // #568
+   {// special case: entries consisting of a single space sort first!              // #568
+    if (str1.charAt(0) == ' ') return 1;  // blank entry always smaller            // #568
+    if (str2.charAt(0) == ' ') return -1; // blank entry always smaller            // #568
```
jyganci, line 2435: What if both strings are blank?

src/az390.java
```diff
@@ -2410,6 +2421,41 @@ else if (tz390.op_trace_type[index]==613)                          // #533
...
+    for (int i=0; i<7; i++)                                                        // #568
```
jyganci, line 2436: I assume that you know both strings have length at least 7.

src/az390.java
```diff
@@ -2410,6 +2421,41 @@ else if (tz390.op_trace_type[index]==613)                          // #533
...
+     {char c1 = str1.charAt(i);                                                    // #568
+      char c2 = str2.charAt(i);                                                    // #568
+      if (c1 == c2) {} // continue on to next char                                 // #568
+      else // we found different characters                                        // #568
```
jyganci: Replace with "if (c1 != c2)" ???

src/az390.java
```diff
@@ -2410,6 +2421,41 @@ else if (tz390.op_trace_type[index]==613)                          // #533
...
+       {if (c1 == ' ') return -1; // Mnemonic 2 is longer                          // #568
+        if (c2 == ' ') return 1;  // Mnemonic 1 is longer                          // #568
```
jyganci, line 2442: Can they be the same length?

src/az390.java
```diff
@@ -2410,6 +2421,41 @@ else if (tz390.op_trace_type[index]==613)                          // #533
...
+         {if (c2 <= '9') return (c1-c2); // compare two digits                     // #568
+          else return 1;                 // digit always larger                    // #568
+          }                                                                        // #568
+        else // c1 is not numeric                                                  // #568
+         {if (c2 <= '9') return -1; // digit always larger                         // #568
+          else return (c1-c2);      // compare two alphabetic chars                // #568
+          }                                                                        // #568
+        }                                                                          // #568
+      }                                                                            // #568
+    return 0; // First 8 characters are equal                                      // #568
```
jyganci, line 2453: The loop only looked at 7 characters??? (indicies 0-6).

src/az390.java
abekornelis: Then they are the same and the sequence does not matter.

src/az390.java
abekornelis: I did that to allow for the comments. I have put comments on a separate line, and followed your suggestion. 
Looks better this way, indeed. Thanks.

src/az390.java
abekornelis: No, that would mean we have identical mnemonics. Should never occur. If it does we'll have bigger problems than this piece of compare/sort logic.

src/az390.java
abekornelis: I updated the stop criterion. Apparently - for available test cases - it was good enough. But you are correct: it should have inspect all 8 characters of the mnemonic.

src/az390.java
jyganci: If line 2434 is trun, code returns 1, not checking whether str2.charAt(0) is blank. Seems to me that code should return 0 in this case.

abekornelis: As the comments indicate, a blank entry is by definition smaller. Even if they both are blank, equality does not matter, we simply declare the first blank we encounter the smaller of the two. Since the entries are discarded after sorting, the order of the blank entries is irrelevant, as long as they do sort up front.
I could add code to detect equality of entries staged for suppression, but that's more code and - to my mind - more confusing. If you still are opined differently, I'll make the change, though. As stated - it's irrelevant. What's more readable is a matter of style and taste :-)

jyganci (approved):

PR #575, opened 2024-09-23
Add NIST test suite for Cobol

### LINKED ISSUE #574: Add NIST test suite for Cobol to z390 repo
Please add the sources of the NIST test suite for Cobol to the z390 repo.
These sources were used to validate the zCobol functionality.
zCobol does not pass all tests, but at least we have the test code.

### LINKED ISSUE #578: zCobol compiles incur "actr limit exceeded" errors
Assembly of generated bal code ends RC=12 with error message:
MZ390E error  82    (19/916)916 actr limit exceeded

This happens on fields defined as X(nnn) with an initial value that contains an ampersand character.
The logic is flawed: the ampersands in the literal need to be doubled up for the assembly to work okay, 
but the added ampersand is coded as '&&' which still is two ampersands.
The additional ampersand caused macro WS.MAC to go into a loop.

Please fix the issue by unduplicating any literals inside the macro that contain incorrectly duplicated ampersands within a literal string.

abekornelis:
Resolves #574 
Resolves #578


adelosa:
Hi Abe,

I'm not a supporter of adding this very large code base into z390.. i.e. NIST tests.
Where did you get the test source? What is the copyright status of these works?

My preference would be that these are in a separate repo that can be used when required.

Might be worth a chat to discuss this one. Let me know your thoughts.

adelosa (commented):
Need to discuss before progressing this one

jyganci:
Agree with Anthony. Separate repo seems better to me.

adelosa:
@abekornelis @jyganci 
I think the verdict is in for this - separate repo for the test suite..
I have created https://github.com/z390development/nistcobol85 
This contains the uncompressed z archive file and a python script to split the uncompressed file into individual modules.
If we want to add a test run, the script can clone this repo, generate the modules using the python script and compile and run the tests.
That script (of which your BAT file is part) would be the starting point for this process.
I would of course like to see an equiv bash script as this could then be incorporated into the build process and a report published.. For now, I'll settle for a script that uses the new repo..
Hope this makes sense. Let me know if you want to discuss.

abekornelis:
@adelosa I am retracting this PR. Nist suite is already in a different repo. Other changes have been transferred to smaller PRs that are associated with proper issues describing the problem being addressed.

PR #577, opened 2024-09-25
IS576 delete OBJ file when bash/cblcl(g) fail assembly step

jyganci:


abekornelis (approved):
Good catch. Thanks for prompt resolution.

Issue #587, opened 2024-10-03
Compile java without copying sources around


abekornelis:
Mail by John Ganci dd 2024-10-03 18:31

When bash/bldjar was originally written, I questioned why bat/BLDJAR.BAT copied all the java source and the Z390.MAN file to the jar directory. There is a javac option, -d, that specifies the directory to contain the class files. The bash/bldjar script uses that option, which avoids the needless copying of the java source and Z390.MAN. At the time, I chose not to make the corresponding changes to BLDJAR.BAT, thinking Don might've had some reason for doing it the way he did.

Anyway, bash/bldjar works fine. No copying of the java source and Z390.MAN ahead of the compiles.

You might consider changing BLDJAR.BAT to do the same. It is just a small change to the javac and jar commands and simplifying what files are removed from the jar directory before the compiles are done. You'll see what I mean if you look at the two source files side by side.

abekornelis:
Fixed by PR #588

PR #588, opened 2024-10-03
Remove copy steps from bldjar procedure

abekornelis:
Resolves issue #587


bat/BLDJAR.BAT
```diff
@@ -19,24 +19,27 @@ pushd %~dps0..
...
 rem compile java code and create jar file
 if not exist ".\jar\" mkdir jar
-
-erase jar\*.java
-erase jar\*.class
-erase jar\z390.jar
-copy  src\*.java   jar
-copy  src\z390.man jar
 cd jar
+
```
jyganci, line 39: Need to empty the entire jar directory. Either before the cd jar or after. Definitely before the javac. See bash/bldjar to see how it was done for linux.

jyganci (approved):

Issue #573, opened 2024-09-18
Correct OPTABLE(ZS3,LIST) output to match HLASM


abekornelis:
The list of supported mnemonics when running with OPTABLE(Z9) or OPTABLE(ZS3) is incorrect. It should match HLASM's output for the option.

abekornelis:
Fixed - PR#591

abekornelis:
Fixed - PR#593

PR #593, opened 2024-10-28
Fix issue #573 - optable listing for optable(ZS3) should match HLASM

abekornelis:
Fix issue #573 - optable listing for optable(ZS3) should match HLASM

### DIFF
--- a/rt/bat/RUNOPTABLES.BAT
+++ b/rt/bat/RUNOPTABLES.BAT
@@ -16,6 +16,8 @@ set z_MaxRetCode=0
 
 set count_ok=0
 set count_err=0
+set err_optable=
+set err_machine=
 if /I %1. EQU /Q. goto Quick
 if /I %1. NEQ .   goto %1
 :Quick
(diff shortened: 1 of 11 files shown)


src/az390.java
```diff
@@ -2124,11 +2154,11 @@ else if (tz390.op_trace_type[index]==343)
                 case 36:
                     // Alternate formats have mnemonics ending in 'A'
                     if (my_mnemonic.substring(my_mnemonic.length()-1).equals("A"))     // #500
-                       {my_format="RRF";                                               // #500
+                       {my_format="RRF";                                               // #500 #573
```
jyganci, line 2157: Looks the same as the previous line??? Does not appear to be changing anything, but okay.

jyganci (approved):

rt/bat/RUNOPTABLES.BAT
```diff
@@ -116,6 +148,7 @@ call rt\bat\RUNOPTABLE %z_TraceMode% optable UNI *None
 set z_ReturnCode=%ERRORLEVEL%
 if %z_ReturnCode% EQU 0 (set /A count_ok=%count_ok%+1) else (
                          set /A count_err=%count_err%+1
+                         set err_optable=%err_optable% uni
```
jyganci, line 151: Suggestion only. For consistency, change uni to UNI

src/az390.java
abekornelis: I probably made a mistaken change, then changed it back. But left the Issue nr in the comments.. Will remove that.

PR #606, opened 2025-03-29
Correct OPTABLE(ZS4,LIST) output to match HLASM

### LINKED ISSUE #602: Correct OPTABLE(ZS4,LIST) output to match HLASM
The list of supported mnemonics when running with OPTABLE(Z10) or OPTABLE(ZS4) is incorrect. It should match HLASM's output for the option.

abekornelis:
fixes #602


adelosa:
Looks like this change has broken the build. Refer to https://github.com/abekornelis/z390/actions/runs/14145652297/job/39632649478

```
FPCONVRT  MZ390 START
AZ390E abort 139 on line 146          CGIJE   R0,0,ReturnBadParm Null string address: branch
AZ390I invalid opcode type index
AZ390I aborting due to recursive abort error 139 - invalid opcode type index
FPCONVRT  MZ390 ENDED   RC=16 MEM(MB)= MB IO=13
asm ERROR: mz390 rc=16; see errors in mz390 generated linklib/FPCONVRT.BAL/ERR/PRN file(s) and on console
```

TODO for me: Add a check so that its impossible to merge unless the build is working.

adelosa:
OK.. My summary of this change.. Add test scripts to generate all the available instructions for each optable and machine.
The tests will (eventually) compare the output generated against listings generated on a mainframe.
Hopefully I have understood this correctly.

What I would like to see:
1) A bash version of the runner. I'm happy to add these to your branch if you give me access to your fork. They will be much simpler than your current runners..
2) The current BAT and z390tests generate the instruction codes but don't compare against the TF1 files from the mainframe. Personally I think the best place to do this processing is in z390test. You can use the power of Java/Groovy to compare the outputs and you only need to do it once for Win and Linux. Also, trying to do this in Bash or Bat files will be pretty challenging.
You probably know that I'm not a big lover of OPT files.. just creates more files to manage in the repo.. Personally I would have just added the parms in the tests themselves in z390test..
I would also probably just manage the expected results in the test code rather than in files.

I can assist creating the tests if you need help. In the tests, I would read the TF1 file, parse all the instructions, and then do likewise on the generated output. I'd then compare the lists and spit out any differences. Any differences would result in a test failure. 

If you just want to get this merged, then this work can be put on the back burner.

adelosa:
Here is a patch with a basic implementation of bash scripts for this request.
[Basic_runoptables_for_bash.patch](https://github.com/user-attachments/files/19534625/Basic_runoptables_for_bash.patch)

adelosa (changes requested):
See comments in discussion.

abekornelis:
@adelosa Anthony, the generated optable report *is* compared against the TF1 file. But that code lives in the OPTB# program which is executed as part of each test. I have built a test for each optable and machine option that we support. This PR is one in a long series intended to create exactly the validations you are requesting.
As far as bash script is concerned, between John and me there was an understanding that we would create the bash script after the whole effort (of validating optable/machine option results) is completed, so we know we are converting the final version of the bat file, rather than an intermediate one that will need to be kept in sync with every additional change I make while progressing through the list of option values.
And for the OPT files - I use them so that I need only a single .mlc program to run all the tests.
The naming has been constructed to make sure the .opt, .tf1, and output files from assembly all stick nicely together.
And thanks for the patch - I'll keep it for future reference. I'm hoping to go through the remaining option values quickly and complete the whole thing before the summer holidays.

adelosa (approved):
Hi Abe - I am approving based on your comments.. @jyganci will be available for a meeting this week if you wanted to discuss further.

PR #607, opened 2025-03-30
Various Build fixes and enhancements

adelosa:
* Fix issue with relver FORCE parameter not being passed from build.sh so you could not build without git.
* Moved build to Java 17, but using --release 8 meaning binary will work back to Java 8.
* Add support for latest Gradle 8.13 so no issues compiling with latest Java version.
* Add pull request checks to ensure build is working before merge (actions)


abekornelis (approved):
looks okay to me

PR #610, opened 2025-04-13
Creating optable for ZS5, including regression tests

### LINKED ISSUE #612: Correct OPTABLE(ZS5,LIST) output to match HLASM
The list of supported mnemonics when running with OPTABLE(Z11) or OPTABLE(ZS5) is incorrect. It should match HLASM's output for the option.

abekornelis:
fixes Issue #612


src/az390.java
```diff
@@ -2177,6 +2184,26 @@ else if (tz390.op_trace_type[index]==343)
...
+                    else if (tz390.op_trace_type[index]==141)                          // #612
+                       {if (tz390.op_name[index].equals("LOCGR"))                      // #612
+                           {my_format="RRF";                                           // #612
+                            my_operands="R1,R2,M3";                                    // #612
+                            }                                                          // #612
+                        else                                                           // #612
+                           {my_format="RRF";                                           // #612
+                            my_operands="R1,R2";                                       // #612
+                            }                                                          // #612
+                        }                                                              // #612
```
jyganci, line 2196: Suggestion only. Lines 2189 and 2193 could be replaced by one line put either just before or just after the "if else".

src/az390.java
```diff
@@ -2177,6 +2184,26 @@ else if (tz390.op_trace_type[index]==343)
...
+                    else if (tz390.op_trace_type[index]==142)                          // #612
+                       {if (tz390.op_name[index].equals("LOCR"))                       // #612
+                           {my_format="RRF";                                           // #612
+                            my_operands="R1,R2,M3";                                    // #612
+                            }                                                          // #612
+                        else                                                           // #612
+                           {my_format="RRF";                                           // #612
+                            my_operands="R1,R2";                                       // #612
+                            }                                                          // #612
```
jyganci, line 2205: Suggestion only. lines 2199 and 2203 could be replaced by one line put either just before or just after the "if else".

src/az390.java
jyganci, line 2196: Suggestion only. Lines 2189 and 2193 could be replaced by one line put either just before or just after the "if else".

abekornelis: John,
you are correct - that could well be done. However, I am trying to follow a pattern in this silly piece of logic: first go by op_trace_type, then by instruction. It's not been done consistently, I know. But I'm trying. And in due time, all of this silliness should be replaced by something much simpler and more straightforward / elegant. Moving the repeated assignment to my_format out of the if construct would be technically correct, but harder to read/maintain for my snailbrain.

src/az390.java
abekornelis: Same as above. Technically correct suggestion. But I prefer to keep my_format and my_operands assignements together. For me personally that is easier to read, understand & maintain.

jyganci (approved):
Made a few suggestions. Nothing wrong. Assume you will not make the suggested changes. Going ahead and approving/

src/az390.java
jyganci: As I stated in my approval comment, it's okay with me to leave the code as is.

PR #621, opened 2025-04-27
Add support for optable(ZS7) and synonyms, plus test cases

### LINKED ISSUE #611: TestOAttrib groovy script writes output to log
The TESTOAttrib groovy script writes output to the general log stream. It causes a lot of clutter. Can we please redirect the output to go to a file instead?

### LINKED ISSUE #614: Correct OPTABLE(ZS7,LIST) output to match HLASM
The list of supported mnemonics when running with OPTABLE(Z13) or OPTABLE(ZS7) is incorrect. It should match HLASM's output for the option.

### LINKED ISSUE #620: bash scripts in clone have incorrect line endings
When cloning the repository to a Windows environment, line endings for some bash scripts are converted to CR-LF.
This causes problem when running the scripts under wsl (Windows System for Linux)

Please change the .gitattributes file to make sure all bash scripts in any directory are set to have correct line endings.

abekornelis:
Fixes #614
Fixes #611 
Fixes #620 

I've added support for optable(ZS7) and its synonyms, and also added test cases.
The code in az390 for formatting the LIST output of the mnemonics is not pretty (to say the very least).
But it does the job - that's what counts for now.


src/az390.java
```diff
@@ -2102,10 +2103,32 @@ else if (tz390.op_trace_type[index]==202)                          //      #613
                 case 23:
                     my_format="RIE";                                                   // #500
                     my_operands="R1,R3,RI2";                                           // #500 #561
+                    if (tz390.op_trace_type[index]==230)                               // #614
```
jyganci, line 2106: Magic number 230. Many more such follow. Leave for now but make note to create constants???

jyganci (approved):

PR #623, opened 2025-05-01
Fix timeouts during test on slow environment

### LINKED ISSUE #622: Some tests fail on slower machines
Running the z390test regression suite under wsl (linux under windows) is slow enough to cause the default TIME(15) option to be exceeded by some tests. Currently the problem has surfaced only with some SORT tests. Please increase the TIME limit to avoid S422 timer abends, resulting in the test being flagged as Failed. Which is technically correct, but not at all what we really want to test.

abekornelis:
fixes #622


adelosa (approved):
No issues with this although I am surprised your new machine is not running a sort process in 45 seconds.

PR #626, opened 2025-05-02
Add Assist program SOLP06 to regression test suite

### LINKED ISSUE #625: Test program assist\test\SOLP06 not called from RUNASSIST test scripts
The assist feature test program SOLP06 is not called from:
bat\RUNASSISN/BAT
bash\runassist
z390test\...\RunAssist.groovy

Please this test to all three scripts.

abekornelis:
fixes #625


adelosa (commented):
OK, here is a question. One of the reasons that we created the Groovy test framework was to allow us to run the tests in Windows or Unix type envs without having to create specific scripts each time.. but now we are doing it 3 times instead of 2.
What is the value of having OS specific scripts AND a test framework that does the same thing in an OS agnostic way.. Maybe a topic for the next dev meeting that I need to setup. My comment relates to the use of a file containing the comparison which makes sense when you use it 3 times, but not if you just used the test framework.

adelosa (approved):
Approving based on the teams decision that we continue maintaining bash/bat and the test framework.

PR #629, opened 2025-05-10
Move from git tags to version.txt for version management

adelosa:
At our last meeting (May25) a decision was made to move to a git free build process.

The current process relies on git tags, and people who are not familiar with or not using git encounter issues trying to build the package. There have also been some gotchas along the way in terms of making sure git tags are up to date and available for build processes.

The new process relies on a `version.txt` file in the root of the project. This is much simpler approach!

`build.sh` and all subsequent bash scripts are now 100% git free 👏

The release process which still relies on git has been modified to look for changes to the `version.txt` file and if changes are detected, it will trigger a release to be automatically generated.

This request also includes a 'bash/bumpver' which CAN be used to increment the version and commit the version update to the repository. 

Instructions for using bumpver and generating a release have been added to the contributing docs. 
See <https://adelosa.github.io/z390/contributing/contribute_code/#creating-a-new-release>

We will continue to maintain git tags for the project, but these will be automatically created when a release is generated.


adelosa:
Wrong branch.. trying again

PR #635, opened 2025-05-17
Save test output if tests fail to aid diagnostics

adelosa:
Currently when the GitHub tests fail, there is no way to see the output of the tests that failed.

This PR will save the test job output as an GitHub actions artifact if the tests fail.
The test output will then be available to download from GitHub as part of the actions output.
Artifact name is `test-output`

image

New logic has been added to perform printOutput at the end of every test (no need to add to your tests).
The output from tests will be saved to file /z390Test/build/z390test-output.txt.

The change to automatically perform printOutput function at the end of each test means that current usage of this function is often not required at the end of a test. The code ignores when the function is called with no parameters.

If you want to get a snapshot of the output in the middle of a test, then you can still call printOutput - just make sure you include a label.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,6 +58,12 @@ jobs:
           Z390_PROJECT_ROOT="../dist/z390_${VERSION}" z390test/gradlew -p z390test test 
         env:
           VERSION: ${{ needs.build-distribution.outputs.VERSION }}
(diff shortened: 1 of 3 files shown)


abekornelis (approved):
That is a fairly limited amount of changes. I must admit I do not understand the change.
Since it's part of the build process, it cannot break code that z390 users use. I trust the change is okay.

PR #637, opened 2025-05-18
Re-establish tests in demo directory

### LINKED ISSUE #636: Re-establish regression test in demo directory
In the demo directory a series of programs exist that were intended as demo/test programs.
They should be part of the regression test suite, but currently only 4 of them are used.
Also, DEMOWTO1 is a duplicate of the HELLO program and should be deleted.

### LINKED ISSUE #643: File compares in RunAsmDemos.bat are failing in pre-built distribution
The script in BAT\RUNASMDEMOS.BAT contains two file comparison steps to validate correct operation of test programs.
When running this script (on Windows) using a pre-built distribution of z390 (which is created on Linux) the reference file contains Linux-type line endings (LF only) whereas the created output file contains the proper line endings for the host operating system (Windows for running a BAT script, making the line endings CR-LF). This causes the file comparison to fail.

Please find a solution to re-enable these comparison tests as part of z390 demo/validation.

abekornelis:
Fixes #636 
Fixes #643


bat/RUNASMDEMOS.BAT
```diff
@@ -11,46 +11,53 @@ if /I "%1" == "tron" (set z_TraceMode=tron
...
-::if %z_ReturnCode% GTR 0 (set z_ReturnCode=8
-::                         echo %0 ERROR: TESTDCB1 data were not copied correctly
-::                         goto return
-::
-::                         )
-:: next 4 lines added to temporarily match bash version
-:: type demo\TESTDCB1.TF2
-:: echo "Verify demo\TESTDCB1 copied demo\TESTDCB1.TF1 to demo\TESTDCB1.TF2"
-:: echo "Verify demo\TESTDCB1 wrote to demo\TESTDCB1.TF3"
+                      call bat\ASMLG %z_TraceMode% demo\TESTDCB1 %1 %2 %3 %4 %5 %6 %7 %8 %9 || goto error
```
jyganci: Line appears to be indented???

bat/RUNASMDEMOS.BAT
```diff
@@ -11,46 +11,53 @@ if /I "%1" == "tron" (set z_TraceMode=tron
...
-::if %z_ReturnCode% GTR 0 (set z_ReturnCode=8
-::                         echo %0 ERROR: TESTDCB2 data were not copied correctly
-::                         goto return
-::
-::                         )
-:: next 4 lines added to temporarily match bash version
-:: type demo\TESTDCB2.TF2
-:: echo "Verify demo\TESTDCB2 copied demo\TESTDCB2.TF1 to demo\TESTDCB1.TF2"
-:: echo "Verify demo\TESTDCB2 wrote to demo\TESTDCB2.TF3"
+                      call bat\ASMLG %z_TraceMode% demo\TESTDCB2 %1 %2 %3 %4 %5 %6 %7 %8 %9 || goto error
```
jyganci: Line appears to be indented???

jyganci (commented):
Looks okay. Made a couple  of comments.I'll let Anthony approve.

bat/RUNASMDEMOS.BAT
abekornelis: I aligned with the call statement on the next line - they belong together as a pair.

bat/RUNASMDEMOS.BAT
abekornelis: As above, I aligned with the call statement on the next line - they belong together as a pair.

bat/RUNASMDEMOS.BAT
```diff
@@ -11,46 +11,53 @@ if /I "%1" == "tron" (set z_TraceMode=tron
 set /A z_NestLevel=%z_NestLevel%+1
 rem ----- Lvl(%z_NestLevel%) Start %0 %1 %2 %3 %4 %5 %6 %7 %8 %9
 
+rem during build on github server the compare steps fail - no clue why.
```
adelosa: Do you want me to investigate reason for fail when running on GitHub?

adelosa (commented):
Besides the test failing on GitHub don't have issue. For that we either fix or remove. At this point I have no idea if the test only works on your machine. Nothing special about GitHub runners so good chance won't work on others. Let me know if you want me to work out the issue.

bat/RUNASMDEMOS.BAT
abekornelis: With PR#635 in place, I should be able to diagnose the issue myself. Will have a look when time permits.

adelosa (commented):
Looks like the test is still being bypassed for GitHub. I actually know what the issue is and it will impact all users of the distribution (I.e not cloning the repo). The distribution is built on Linux which means that the file you compare will have Linux file endings. The one generated will have Windows file endings. Hence the failure. You need to work out how to compare ignoring line endings. There may be some comp options you can use for this.

abekornelis:
@adelosa You do have a point there. Although the comparison has been disabled for a long time, it should be re-enabled. The problem is that neither comp nor fc utility support an option to ignore line endings. So we must fall back to copying the reference file using copy or xcopy, but neither of those has an option to force a specific type of line ending. I found some tools to convert the line endings but nothing native to Windows. I am for now at a loss and do not have a solution other than writing our own comparison program. I think that kinda beats the purpose.

adelosa:
I think this is returning to my point about the purpose of bat/bash vs unit tests. This test is very easy to do using groovy/java and junit. They are designed for writing tests. I don't think you can say the same about windows batch commands. 
I'm sure ai will have the answer.

bat/RUNASMDEMOS.BAT
```diff
@@ -12,45 +12,48 @@ set /A z_NestLevel=%z_NestLevel%+1
...
+call bat\ASMLG %z_TraceMode% demo\DEMODFP1 nocon %1 %2 %3 %4 %5 %6 %7 %8 %9 || goto error
+call bat\ASM   %z_TraceMode% demo\DEMONUM1 nocon %1 %2 %3 %4 %5 %6 %7 %8 %9 || goto error
+call bat\ASMLG %z_TraceMode% demo\DEMONUM2 nocon %1 %2 %3 %4 %5 %6 %7 %8 %9 || goto error
+call bat\ASMLG %z_TraceMode% demo\DEMOSTR1 nocon %1 %2 %3 %4 %5 %6 %7 %8 %9 || goto error
+call bat\ASMLG %z_TraceMode% demo\DEMOWTO2 nocon %1 %2 %3 %4 %5 %6 %7 %8 %9 || goto error
+call bat\ASMLG %z_TraceMode% demo\SIEVE    nocon %1 %2 %3 %4 %5 %6 %7 %8 %9 || goto error
+call bat\ASMLG %z_TraceMode% demo\STDDEVLB nocon %1 %2 %3 %4 %5 %6 %7 %8 %9 || goto error
+
+rem when running this script using a pre-built distribution, the line endings are Linux-type causing file compares to fail
+rem we've disabled the file compares as of May 24, 2022. Issue 643 covers this case and requests a repair.
```
jyganci, line 28: May 24, 2022 should be May 24, 2025 ???

abekornelis: No, the stated date is correct. That date PR#418 was merged which disabled the comparison steps.

adelosa (dismissed):
I pushed an update to the branch that reinstated the file compare on Windows. It strips the line endings using findstr before the compare. If you are happy with the update, then happy for you to merge.

bat/RUNASMDEMOS.BAT
```diff
@@ -30,14 +30,26 @@ rem we've disabled the file compares as of May 24, 2022. Issue 643 covers this c
...
-    call bat\ASMLG %z_TraceMode% demo\TESTDCB1 %1 %2 %3 %4 %5 %6 %7 %8 %9 || goto error
-rem call comp.exe demo\TESTDCB1.TF1 demo\TESTDCB1.TF2 /A /M               || goto error_dcb1
+call bat\ASMLG %z_TraceMode% demo\TESTDCB1 %1 %2 %3 %4 %5 %6 %7 %8 %9 || goto error
+
+REM Normalize line endings for comparison (remove CRs) using only batch commands
+if exist "%TEMP%\dcb1_1.tmp" del "%TEMP%\dcb1_1.tmp"
+if exist "%TEMP%\dcb1_2.tmp" del "%TEMP%\dcb1_2.tmp"
+findstr /V /R "" demo\TESTDCB1.TF1 > "%TEMP%\dcb1_1.tmp" 2>nul
+findstr /V /R "" demo\TESTDCB1.TF2 > "%TEMP%\dcb1_2.tmp" 2>nul
+comp.exe "%TEMP%\dcb1_1.tmp" "%TEMP%\dcb1_2.tmp" /A /M || goto error_dcb1
```
jyganci: just a question. The original was "call comp.exe ...". You've replaced it with "comp.exe ...". Is this intended?

abekornelis: Using explicit call instead of implicit, and adding file extension, is a habit of mine that has grown from precaution. It avoids invoking the 'wrong' thing when run on a system that I cannot control. Paranoid? I guess so :-) 
I did put the call back in, and I removed the delete steps, the redirected output using a single `>` will "create-or-replace" the target file anyway. (You can use `>>' to "create-or-append" but that's not what we want in this case)
A big thank-you to Anthony for providing this solution. Nifty, but effective and straightforward. I like that!

adelosa (approved):
Looks good to me

jyganci (approved):

PR #644, opened 2025-05-31
Improve optable definitions and processing

### LINKED ISSUE #631: Improve optable definitions in z390
- make names of optables consistent
- add comments to associate each optable with the defining Principles of Operation
- modify the code in tz390.create_opcodes() to get rid of the long and repetitive lists of hardcoded optable names, spelled out for each optable option value.
- provide bash script equivalent to rt\bat\RUNOPTABLES.BAT

abekornelis:
Fixes #631


src/tz390.java
```diff
@@ -1125,7 +1126,27 @@ public  class  tz390 {
     int[]    optable_option_nr = null;                    // #554
     String[] optable_option_id = null;                    // #503
     String[] optables_optable = null; // optable's optable// #503
-    String[] optable_optable_equivalence =                // #503
+    public static final int OPCODE_FOR_360_20 =  0;       // #631
```
adelosa, line 1129: Use an enum here?

doc/user_guide/z390/z390_options.md
adelosa, line 1: Not sure why the change from * to - here. In markdown, there is no difference (will be rendered as an unordered list) but the general style of the docs uses asterix (*).

rt/bat/RUNOPTABLES.BAT
adelosa, line 1: I can't comment on this as I don't use Windows and I hate windows batch for anything that is more than trivial. I'll take your word for this that it works as intended.

src/az390.java
adelosa, line 1: see comment below re: use of an enum rather than just a series of constants

src/az390.java
```diff
@@ -2080,7 +2081,7 @@ else if (tz390.op_name[index].equals("MVCK")
                 case 20:
                     my_format="RSE";                                                   // #500 #554
                     my_operands="D1(B1),D2(B2)";                                       // #500
-                    if(tz390.opt_optable_optb_nr >= 6) // YOP ff. change to RSY        // #568
+                    if(tz390.opt_optable_optb_nr >= tz390.OPCODE_FOR_YOP) // YOP ff. change to RSY // #568 #631
```
adelosa, line 2084: OK, confused now.. Its like a constant, but you are doing a numerical compare.. I still think opcode to keep the codes logically grouped.

adelosa (commented):
Can you see if its possible to refactor as an enum for the Arch level. Here is some advice from AI on this. If you get stuck, I can have a look.

# AI feedback
You're correct—using an enum instead of straight integer constants for the architecture levels (e.g., OPCODE_FOR_ESA, OPCODE_FOR_Z9, OPCODE_FOR_Z12, etc.) would improve code readability, maintainability, and type safety.

**Benefits of using an enum:**
- **Clarity:** Code like tz390.opt_optable_optb_nr >= tz390.OPCODE_FOR_Z9 is less clear than tz390.opt_optable_optb_nr >= OpcodeArchLevel.Z9.
- **Grouping:** All related values are grouped together in one place, making it easier to see the valid set of options.
- **Type Safety:** Enums prevent accidental assignment of unrelated values.
- **Extensibility:** Adding, removing, or renaming architecture levels is simpler and less error-prone.

**Example of how an enum could look in Java:**
```java
public enum OpcodeArchLevel {
    ESA(4),
    ZOP(5),
    YOP(6),
    Z9(7),
    Z12(10),
    Z14(12),
    Z15(13),
    Z16(14);
    // ... add more as needed

    private final int value;
    OpcodeArchLevel(int value) {
        this.value = value;
    }
    public int getValue() {
        return value;
    }
}
```
**Usage in code:**
```java
if (tz390.opt_optable_optb_nr >= OpcodeArchLevel.Z9.getValue()) {
    // ...
}
```

**Considerations:**
- You'd need to refactor references and likely update serialization, configuration, or UI code if these values are stored or displayed.
- If you ever want to switch to a more descriptive or complex representation (not just ints), enums make this transition easier.

**Summary:**  
Refactoring the architecture level constants to an enum is a best practice in modern Java and would benefit the long-term health of the codebase. If you want, I can help draft what a full refactor might look like.

doc/user_guide/z390/z390_options.md
abekornelis: that was done for consistency with other unnumbered lists further down in the document.

src/tz390.java
abekornelis: Rather not - only a few are assigned arbitrary numbers, or codes as you wish. Most of these numbers are assigned to define a sequence of events. Being able to do a meaningful numeric comparison is essential.

adelosa (approved):
This PR could be doing more to improve the code base generally but it does improve the codebase a little. I provide my approval on that basis and in the interests of progression.

PR #657, opened 2025-07-28
Fix javadoc comments for az390

abekornelis:
Creating javadoc with the latest compiler I get tons of warnings.
I have converted existed plain comments to javadoc comments adding @param and @return where needed.
In some places I have also modified whitespace and/or html markers to improve formatting.

### DIFF
--- a/doc/contributing/contribute_docs.md
+++ b/doc/contributing/contribute_docs.md
@@ -14,6 +14,9 @@ Markdown uses simple text files with specific syntax for formatting. See [the ma
 
 You can also use functions provided by the admonition extension in mkdocs. See the [doco for admonition](https://python-markdown.github.io/extensions/admonition/)
 
+**Note:** As part of the normal build process the documentation library for the Java code is generated from the java source code using the standard java tool javadoc.
+Generated documents can be found in z390 subdirectory build\javadoc or build/javadoc depending on your Operating System. These documents are *not* published to the web.
+
 ### Structure
 
 All the documentation is contained in the `doc` folder.
--- a/src/az390.java
+++ b/src/az390.java
@@ -43,6 +43,9 @@ import java.util.regex.Pattern;
 
 import javax.swing.JTextArea;
 
+/**
+  * Class az390 implements the assembler part of z390
+  */
 public  class  az390 implements Runnable {
    /*
     az390 is the assembler component of z390 which can be called from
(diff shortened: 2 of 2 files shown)


adelosa (commented):
Can you please format the module using an IDE or formatter so that the comments line up with the code. Also, I know that you are getting rid of javadoc warnings, but I'm not sure of the value of adding comments like "variable".. Personally I think the Javadoc warning is a placeholder to provide a real comment when ready but not expecting you to change this. I quickly reviewed the generated Javadocs and found some issues. For example, you added dashes after the function params. [Oracle Javadoc guide](https://www.oracle.com/au/technical-resources/articles/java/javadoc-tool.html#styleguide) says the following "Additional spaces can be inserted between the name and description so that the descriptions line up in a block. Dashes or other punctuation should not be inserted before the description, as the Javadoc tool inserts one dash." Just saying, have a look at the generated docs and make sure it looks OK.
Can you have a look at these items.

abekornelis:
Anthony, thanks for the suggestions. I have made another round of changes. I think it looks better now. 
I'm aware there's still a lot of content missing. That's for another time. This is about as much as I can do in this round of improvements.

adelosa:
Ok. Ignoring the code style issue, would you consider aligning the comments with the method names. They are indented like when they were below the method but now they are above the method the indent should be removed. If you do that, I can sign off.

abekornelis:
@adelosa thanks for offering a solution that makes us both feel good. I'll be happy to make those changes when time permits. Thanks!

abekornelis:
@adelosa I have fixed the first few javadoc comments blocks to validate that this is the 'correct' format. Also, the placeholder text "variable" I used for the variables is meaningless. I'm thinking to change that to "Missing description" or some such. Do you have a suggestion?

adelosa:
Comment indent updates look good. No idea what to do for blank comments. I'm ok with them as is.

abekornelis:
Okay, thanks. I'll fix the rest of the comment indents. I'm leaving the placeholder comments for variables as is. Hopefully, over time, doco will improve further.

abekornelis:
I think I've fixed the indentation everywhere it was needed. If you find any more flaws, please let me know. Will be happy to fix.

adelosa (approved):
Looks good to me.

PR #665, opened 2025-08-16
Windows support for z390 Build using Java 8 SDK

abekornelis:
The build procedure supplies `--release 8` as a parameter to the java compiler. This is the correct syntax for Java version 9 and above. Since we're supporting java version 8, the version 8 compiler should also work okay. But it does not support `--release 8`. Instead it requires `-source 8 -target 8`.
This change checks the java version to make sure the correct parameters are passed to request version 8 compatible behaviour.


adelosa:
This PR should be named "Windows support for z390 Build using Java 8 SDK". It's already correct for java 9 and above.

Please review <https://stackoverflow.com/questions/43102787/what-is-the-release-flag-in-the-java-9-compiler>
Indicates that -source and -target also require -bootclasspath option specified to ensure that compiled binary works on older versions.

Docs reference expected java versions for development. <https://z390development.github.io/z390/contributing/contribute_code/#install-java-sdk>

This only targets Windows builds.. is this just a Windows thing to support builds on Java 8?

I could live with separate scripts that say J8 in them and then we can delete once people move to supported stacks!
So bldjarj8 maybe??

abekornelis:
@adelosa :
1. We never used -bootclasspath in the past. I would not know what path to put in. I guess that would depend on where the compiler was installed on the system where the script is executed.
2. I'm okay to write a modification to the requirements section for the java compiler.
3. For now this is just a Windows issue. I know of only one user site that is bound to version 1.8. I doubt that anyone would shout for joy if we carry this over to the bldjar shell script.
4. Having a separate script would require the logic to determine the java version to move into the BUILD.BAT - to my mind that is not an improvement. Knowledge about java should be confined to java-related scripts.

jyganci (approved):

Issue #659, opened 2025-08-01
CCW/CCW0/CCW1 with location counter reference in address field generates incorrect object code.


jbs1220:
Summary: The z390 assembler generates incorrect object code for the
address field within a CCW/CCW0/CCW1 when the address expression
refers to the location counter.

When an address relative to the location counter -- i.e., an address
expression including an asterisk -- is the 2nd positional parameter
of a CCW/CCW0, HLASM uses the first byte (byte 0) of the CCW as the
value of the location counter when computing the resulting address.
However, the z390 assembler uses the first byte of the address
expression -- i.e., byte 1 (the 2nd ordinal byte) -- of the CCW as
the value of the location counter. The result is that z390 assembles
a CCW address field value that is 1 greater than the value that HLASM
assembles.

Similarly, in the case of a CCW1, the z390 assembler uses byte 4 of
the generated CCW as the value of the location counter when
generating the address field of a CCW1, resulting in an address field
that is 4 greater than the value that HLASM assembles.

The assembler should not update the location counter until it has
finished assembling the entire CCW/CCW0/CCW1 assembler instruction.

The HLASM Language Reference manual (SC26-4940-09), p. 32, says that
while assembling a machine or assembler instruction, the location
counter is "the value assigned to the symbol, if present, in the name
field of the statement." That is the address of the leftmost byte
(byte 0) of the CCW/CCW0 or CCW1.

(I realize that this issue may seem academic, since the z390 emulator
does not support executing channel programs. However, if the
assembler is intended to support all of the assembler instructions
that HLASM does, it should generate the correct object code.)

Environment:

    z390 rev:   1.8.3
    Java:       Java 8 Update 461

Source file CCWLOC.MLC:

    CCWLOC   CSECT ,
    *
    * CCW COMMANDS:
    CCWCTIC  EQU   X'08'    TIC
    CCWCSEEK EQU   X'07'    SEEK
    CCWCSIDE EQU   X'31'    SEARCH ID EQUAL
    CCWCRDAT EQU   X'06'    READ DATA
    *
    * CCW FLAGS:
    CCWFCD   EQU   X'80'    CHAIN DATA
    CCWFCC   EQU   X'40'    CHAIN COMMAND
    CCWFSLI  EQU   X'20'    SUPPRESS INCORRECT LENGTH INDICATION
    CCWFSKIP EQU   X'10'    SKIP DATA TRANSFER ON RD/RDBK/SENSE
[...]

PR #677, opened 2025-09-27
Fix CCW/CCW0/CCW1 with location counter reference in address field problem for issue 659

jyganci:
The az390.java code incorrectly processes the code generation for CCW/CCW0/CCW1 due to updating the location counter value as it parses the CCW parameters. The HLASM LR states that the location counter remains constant when processing one statement.

The problem only manifests itself when a location counter reference is used for the data address, which is positioned at CCW/CCW0 offset 1, CCW1 offset 4. The az390.java code adds 1, respectively 4, before processing a data address value of the form "*-8", resulting in an incorrect value being generated.

The az390.java code has been updated to use a location counter offset value when generating relocatable expression values. The offset is 0 in most cases, maintaining existing processing for other than CCWs. The az390.java methods gen_ccw0() and gen_ccw1() have been modifed to use offsets of 1, respectively 4, when generating the data address values, leaving the location counter value unchanged.


abekornelis:
If this PR fixes issue 659, please add a line at the top of the first (initial) comment:
Fixes #659 

Also, please fix any javadoc issues - the current build.bat fails due to these. I'll make the changes locally and will be happy to share them if you want me to.

Issue #660, opened 2025-08-07
Incorrect displacement generated in listing and object code.


DanGreiner:
Using the z390 assembler version 1.8.3, I'm attempting to assemble an old S/360 diagnostic program. The object code (in both the listing and object file) show an incorrect location for certain variables, even though the symbol listing shows them correctly.

A sample is shown below:

```
001010 05E0                                 (1/55)51 RTN02    BALR  R14,R0                   Save routine starting address.
001012 41400000                             (1/56)52          LA    R4,0                CLEAR REGISTER 4
001016 5850F110                001110       (1/57)53          L     R5,DVDND2
00101A 5860F110                001110       (1/58)54          L     R6,DIVSR2
00101E 1D46                                 (1/59)55          DR    R4,R6               DIVIDE
```
Note that DVND2 (the lease significant bits of a dividend) and DIVSR2 (a divisor) are at two different locations (1110 and 1114, respectively), but the code generated shows the wrong location for DIVSR2. However, the symbol listing correctly shows DIVSR2 at location 1114.

The attached source file can be assembled without any additional macros (you'll have to rename its file type, since GitHub doesn't like .asm or .mlc). The assembled listing will show five occurrences of incorrect locations. I've run into similar problems from faulty symbol hashing, but that doesn't explain why the symbol listing has correct locations.

[badcode.txt](https://github.com/user-attachments/files/21650430/badcode.txt)

DanGreiner:
The following program illustrates the problem more concisely:  The DC statements that contain the addition of two (or more) symbols (e.g., ADDR2, ADDR4, and ADDR8) results in incorrect locations for later symbols when referenced by the code. 
```
Assembler Listing
000000                                        (1/1)1 CRAPOLA  CSECT 
000000                                        (1/2)2          USING CRAPOLA,15
LISTUSE CRAPOLA  ESD=0001 LOC=00000000 LEN=01000 REG=F OFF=00000 LAB=
000000 5810F108                000108         (1/3)3          L     1,ADDR1
000004 5820F10C                00010C         (1/4)4          L     2,ADDR2
000008 5830F10C                00010C         (1/5)5          L     3,ADDR3
00000C 5840F110                000110         (1/6)6          L     4,ADDR4
000010 5850F110                000110         (1/7)7          L     5,ADDR5
000014 5860F114                000114         (1/8)8          L     6,ADDR6
000018 5870F118                000118         (1/9)9          L     7,ADDR7
[...]

PR #683, opened 2025-10-06
Fixes for Incorrect displacement generated in listing and objcode #660

jyganci:
When creating A-cons for complex relocatable expressions, RX instructions that reference such a field have incorrect displacement values due to the location counter value for the complex relocatable expression A-con not being incremented. See sample code in issue #660.


abekornelis (approved):
Passes all my tests.

PR #687, opened 2025-10-11
Convert constants for architecture levels to enum

### LINKED ISSUE #656: Improve code for defining architecture levels
Architecture levels are defined using constants in tz390.
Please replace with an enum. For details, please see (closed) PR 644.

abekornelis:
Fixes #656 

Convert constants for architecture levels to enum

there is a little problem: I cannot figure out how to make the references in az390 to the enum elements compile.
I do not understand what the compiler is complaining about :-(
Could you please suggest how to resolve this compile issue?


abekornelis:
The java compiler is unable to differentiate between the class tz390 and the instance tz390 declared in az390.
This causes compile errors that I am unable to resolve.

adelosa:
Seems too many things are called tz390. I created a new file containing the enum and referenced that in tz390 and az390 and it compiles and runs almost fine. Seems that rt test TESTERR2 is now returning RC12 instead of RC16. I don't see what would cause this change in return behavior. You may have a better idea about that one. 
General convention in Java is new class/enum, new file. 
Embedded classes and enum are not normal Java coding conventions, especially when you want to reference it outside of the scope of the embedded class. If its only referenced within the parent class, then this is OK.
I created a branch in my fork with the changes - See https://github.com/adelosa/z390/tree/optables. 
Let me know your thoughts.

abekornelis:
@adelosa thanks - I was able to copy the changes from your fork. The program now compiles okay and passes all tests. I checked the change for TESTERR2 RC going from 16 to 12. The 16 actually indicated an undesired error. The errors tested in TESTERR2 are supposed to cause RC=12. I adjusted the expectation in the groovy script.

There is only one - minor - issue left. The javadoc parser (at least for my version of java) does not recognize the javadoc comments preceding the enum in the new ArchLevel.java - I have not been able to find the cause of this. It feels a bit awkward to introduce a new warning while I've been working hard to get rid of all those warnings. Could you please have another look at it? Maybe you can spot what I'm missing.

It's not a blocker, as far as I'm concerned. If you feel like it and can spare the time, I'd appreciate a review or an approval.

adelosa (approved):
LGTM ✅

PR #690, opened 2025-10-11
Make zCobol ignore comment-only paragraphs

### LINKED ISSUE #655: zCobol fails on number in INSTALLATION paragraph of Identification division
The INSTALLATION paragraph is supposed to contain comments only, until the next paragraph or division is encountered.
ZC390 seems (to be verified) to stop at the first period. this causes a subsequent line that starts with a number to be interpreted as a working storage definition. Which causes the error:
`ZC390E zcobol invalid ws level 5203 on line 11`

### LINKED ISSUE #689: zCobol fails to detect out-of-order sections
zCobol assemblies fail to flag as error when prescribed DIVISIONS, SECTIONS, and paragraphs appear out of order or repeatedly.

Please enhance the zCobol compiler to flag out-of-order or duplicates as error.

abekornelis:
fixes #655 
fixes #689 

I have tested this with the standard set of test cases, plus the full zCics build
The proper test case is the program where the issue first surfaced: the NIST suite test program CM101M.CBL
An abridged version of  that program is included as test case CM101M01 for z390.


src/zc390.java
```diff
@@ -226,17 +227,262 @@ h.  Single () wrapped in single quotes.
...
+             };                                                         // #655
+    static section_definition[] program_sections = null;                // #655
+    /*
+     * end of global data
+     */
+
+
+
+/**                                                                                                                        // #655
+ * Custome exception                                                                                                       // #655
```
Zaneham: This is a test of me using the comment system, but this is following on from the call we had. This is where I noticed the orphan :-)

abekornelis: Hi Zane, I fixed the typo in the comment. But I'm not finding any orphan. The exception is used in various places in the new code. Could you please be a bit more specific what exactly you think is orphaned?

Zaneham (approved):
Approved. I found it funny the office of the org that validated COBOL compilers threw us those errors.

I've re-run the tests and everything passes; the issues are addressed.

PR #700, opened 2025-10-22
Fix actr limit exceeded during assembly phase of zCobol compile

### LINKED ISSUE #578: zCobol compiles incur "actr limit exceeded" errors

abekornelis:
fixes #578


rt/bash/runcbltests
```diff
@@ -14,6 +14,7 @@ bash/cblclg zcobol/tests/TESTIF3 $1 $2 $3 $4 $5 $6 $7 $8 $9
 bash/cblclg zcobol/tests/TESTMOV1 $1 $2 $3 $4 $5 $6 $7 $8 $9
 bash/cblclg zcobol/tests/TESTMOV2 $1 $2 $3 $4 $5 $6 $7 $8 $9
 bash/cblclg zcobol/tests/TESTMOV3 $1 $2 $3 $4 $5 $6 $7 $8 $9
+bash/cblclg zcobol/tests/CM101M03 $1 $2 $3 $4 $5 $6 $7 $8 $9
```
jyganci: In the corresponding bat/RUNCBLTESTS.BAT change and the corresponding groovy script change you did CBLC, not CBLCLG. Reason the bash version is different?

jyganci:
Is there an issue that describes this problem. Sample program that produces the error?

rt/bash/runcbltests
abekornelis: The issue #578 surfaced while trying to compile test programs from the NIST suite. I have created a trimmed-down version of the program to validate correct compilation. The issue occurs in the working storage section. The procedure division has been reduced to just stop run. The program does nothing. It could be run, but there is no added value in doing so.

zcobol/mac/WS.MAC
```diff
@@ -903,11 +904,11 @@
          AEND
          AIF  ('&PIC_TYPE' EQ 'X'                                      X
                AND '&PIC_VAL' NE '')
-               :&AMP SETA ('&PIC_VAL' FIND '&&')
+               :&AMP SETA ('&PIC_VAL' FIND '&')
```
jyganci: FIND('&')  and "&STR:,'&' are invalid when using HLASM. See HLASM LR Chapter 9, ""Evaluatoin of chasracter expressions". Page 341 for my copy.

The original '&&' in WS.MAC for both FIND and concatenation are correct.

There are other problems with the code you added here that I cannot explain yet.

abekornelis: Thanks for pointing that out. I understand I made the situation worse by applying an incorrect fix. Apologies for that.
I am inclined to retract the changes and the PR, and restart from square zero to analyse and fix the infinite loop in the macro? Or do you see an alternative solution?

jyganci: I think I have a fix for the infinite loop. Needs a little more testing and needs to be converted to using the structured macros (I did not use them but the code is structured using AIF and AGO. Will send it to you via email when I'm done.

jyganci (approved):

PR #701, opened 2025-10-22
Changes to change CDE CDUSE field usage from byte to halfword Is658

### LINKED ISSUE #658: CDE field CDEUSE is a halfword but sz390.java treats it as a byte
When creating/deleting/updating a CDE, sz390.java incorrectly manages the CDUSE field. The DSECT for the CDE, generated by macro CDED.MAC, defines the field as a halfword, as does the zOS documentation. However, sz390.java treats the field as a byte. Need to update sz390.java to change

byte[]    cde_use   = (byte[])Array.newInstance(byte.class,max_cde_pgms);

to

short[]    cde_use   = (short[])Array.newInstance(short.class,max_cde_pgms);

and modify all references to cde_use from byte access to short access.

jyganci:
fixes #658 
The z390 CDE macro defines CDUSE as a halfword. The z390 module sz390.java manages CDUSE and its own internal parallel array "byte cde_use[]" as a byte with maximum CDUSE and parallel array values of 255 = X'FF'. Module sz390.java  is modified as follows: (1) internal parallel array changed to "short cde_use[]"; (2) use halfword when referencing CDUSE; and (3) maximum value for both is 32767 = X'7FFF' = maximum positive halfword value. Note: the z390 behavior when CDUSE reaches its maximum value differs from z/OS behavior. See comments in rt/mlc/IS658.MLC.

### DIFF
--- a/z390test/src/test/groovy/org/z390/test/RunAsmTests.groovy
+++ b/z390test/src/test/groovy/org/z390/test/RunAsmTests.groovy
@@ -5,6 +5,8 @@ import org.junit.jupiter.api.Test
 class RunAsmTests extends z390Test {
 
     var options = ['trace', 'noloadhigh', "SYSMAC(${basePath("mac")})", "SYSCPY(${basePath("mac")})"]
+    var optionsNoinit = ['trace', 'noinit', "SYSMAC(${basePath("mac")})", "SYSCPY(${basePath("mac")})"]
+    var optionsNoinitNoloadhigh = ['trace', 'noinit', 'noloadhigh', "SYSMAC(${basePath("mac")})", "SYSCPY(${basePath("mac")})"]
 
     @Test
     void test_TESTINS1() {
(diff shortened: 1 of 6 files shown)


z390test/src/test/groovy/org/z390/test/RunAsmTests.groovy
```diff
@@ -97,6 +99,18 @@ class RunAsmTests extends z390Test {
         assert rc == 0
     }
     @Test
+    void test_IS658() {
+        int rc = this.asml(basePath("rt", "mlc", "IS658LD"), *options, 'optable(z390)')
```
abekornelis, line 103: Just FYI: the optable option is not be needed. z390 is the default. Since it does no harm to be explicit, so you're okay here.

abekornelis (approved):
I updated the PR description to link the PR to the underlying issue.
It makes github close the issue when the PR is merged. 
If you don't want that - just take that line out.
Thanks for making this fix!

z390test/src/test/groovy/org/z390/test/RunAsmTests.groovy
jyganci: Will create an issue to delete the use of optable on relevant rt/bash/runasmtests, rt/bat/RUNASMTESTS.BAT and RunAsmTests..groovy.

PR #703, opened 2025-10-22
remove superfluous optable(z390) Is702

### LINKED ISSUE #702: Remove "optable(z390)" from various tests in rt/bash/runasmtests, rt/bat/RUNASMTESTS.BAT, and RunAsmTests.groovy
Since the default for the "optable" option is "z390", it is not necessary to use it. Okay to remove its use in some of the tests in rt/bash/runasmtests, bat/RUNASMTESTS.BAT, and RunAsmTests.groovy.

jyganci:
fixes #702


abekornelis (approved):
Looks good.
Thanks John.

PR #705, opened 2025-10-26
Fix javadoc comments for pz390

abekornelis:


jyganci:
What is the purpose of the "/** variable */" comments? Is this just a first step in creating javadoc comments for pz390?

abekornelis:
> What is the purpose of the "/** variable */" comments? Is this just a first step in creating javadoc comments for pz390?

We've been here before. No problem. The javadoc tool throws warnings for every global variable without proper comments. It's too much work to provide proper documentation for each and every one. So I'm putting in placeholders. With any subsequent change, if the programmer figures out how a variable is used, they are more than welcome to provide their findings in the comment and remove the placeholder. For now, it's just that. It's meagre, I know - but it's the best I can do at this time. Otherwise the change would take too much time - and it would result in an even bigger delta. This one's sizeable enough, I think.

BobMissing (approved):
Huge changes seem to confuse the source compare a lot.  I got through it anyways.

BobMissing (approved):
Second attempt to approve this pull request.

PR #707, opened 2025-11-03
improve zCics build procedure

abekornelis:
improve zCics build procedure


Zaneham (approved):
Approved- clean refractor of the build scripts as far as I can see.

PR #712, opened 2025-11-07
fix for incomplete execution of windows tests

### LINKED ISSUE #711: Regression test for Windows environment is not executed completely
The RT run is governed by the build.yml script in the github\workflows directory of the repository.
I contains routine run-windows-demo which calls RUNASMDEMOS.BAT and RUNCBLDEMOS.BAT
But it executes only the first one.

Cause: the bat files are called implicitly (i.e. without a CALL verb) which acts (on windows) like a XCTL rather than a CALL.
Hence, when the first bat file ends, it terminates the windows 'cmd'  shell and subsequent commands cannot be executed.

Please make the calls explicit to ensure both procedures get executed.

abekornelis:
fixes #711


adelosa (approved):
LGTM ✅

PR #713, opened 2025-11-08
zCobol test programs TESTCAL1-4 moved for inclusion in gradle test script

abekornelis:
zCobol test programs TESTCALx moved for inclusion in gradle test script


abekornelis:
Still in draft mode. Need to merge #706 first, then resolve the conflicts between that change and this one (if any)

Zaneham (commented):
As far as I can see the tests are only additive and do not impact runtime behaviour. Parity between Bash and BAT is maintained. 

Although I'm new the new cblcl helper seems consistent and helps with testing.

Zaneham (approved):
Approved per previous comment :-) thanks for having me.

PR #725, opened 2025-12-16
Add dummy macro to handle zCobol PURGE verb

### LINKED ISSUE #724: zCobol fails on PURGE verb
The PURGE verb in Cobol is a reserved word, but seems not to be implemented in current IBM Cobol.
The zc390 precompiler correctly processes the verb into a macro, but the macro is missing.
Please provide a sub macro that issues an MNOTE level 8 error message.

abekornelis:
fixes #724 
The bash script for runcbltests was not modified - the test case is expected to fail RC=8
I have no clue how to instruct the bash script to check for RC=8 rather than RC=0


rt/bat/RUNNIST.BAT
Zaneham, line 1: For instructing it to check for rc=8 maybe could try :   
 set +e
  bash/cblc $1/src/CM401M $2 $3 $4 $5 $6 $7 $8 $9
  rc=$?
  set -e
  if [ $rc -ne 8 ]; then
      echo "CM401M: expected RC=8, got RC=$rc"
      exit 1

abekornelis: interesting how you add bash script stuff to a bat file ;-)
This RUNNIST.BAT only reflects the status of the NIST adoption process. I will add comments to make that more clear.

Zaneham (approved):
Approved :-)

Apologies on the late approval.

PR #727, opened 2025-12-19
zCobol verb ALTER not assembling correctly

### LINKED ISSUE #726: zCobol verb ALTER not assembling correctly
The generated code for an ALTER statement contains a flaw. The GEN_ALTER macro generates code to validate that the target of the ALTER is a Jump instruction. If it is not, it abends with an appropriate error message.

The intent is good, but the implementation is flawed: instead of generating a Jump to the ABORT routine (using the ZCVT_ABORT vector in the z390 CVT) with an appropriate error code in R1, it generates a Jump to the error code. Macro line 34 has:
```
         JNE   ZCVT_ABORT_ALTER       NO ABORT
```

Please change the instruction to Jump correctly:
```
         LA     R1,ZCVT_ABORT_ALTER  Error code in case we fail
         JNE   ZCVT_ABORT                  NO ABORT
```

abekornelis:
Fixes #726


Zaneham (approved):
Approved- Ran the tests myself and output shows ALTER working correctly.

Issue #730, opened 2025-12-20
Incorrect code generated for zCobol MULTIPLY verb


abekornelis:
For multiplications where a floating point number is multiplied by a packed decimal, a MP instruction is generated (after conversions). ?Unfortunately, this is wrong. The packed field should be converted, then a floating point multiply should be generated.

Test cases are in TESTCMP1.CBL and in TESTCMP4.CBL
For the time being, these test cases have been commented out. They are marked with **!! and S0C6.
When fixing this issue, please make sure to uncomment the test cases in both test programs.

This change impacts the GEN_MPY macro.

PR #731, opened 2025-12-20
Move zCobol test programs TESTCMP* to tests subdir and reactivate these tests in the test scripts

### LINKED ISSUE #718: For some zCobol multiply operations in an invalid BRAS is generated
The GEN-MPY macro generates BRAS ZC_R14,3 where the three is intended to signal a relative offset of 6 bytes.
The BRAS instruction does not accept a literal value - change to BRAS ZC_R14,*+6 to get correct results.

This case showed up in test program TESTCMP1.CBL

### LINKED ISSUE #719: For some zCobol subtract operations in an invalid BRAS is generated
The GEN_SUB macro generates BRAS ZC_R14,3 where the three is intended to signal a relative offset of 6 bytes.
The BRAS instruction does not accept a literal value - change to BRAS ZC_R14,*+6 to get correct results

This case showed up in test program TESTCMP1.CBL

abekornelis:
Fixes #718
Fixes #719 
Test programs TESTCMP1-6.CBL never made it into the automated test suite.
This PR moves them from the old zcobol/test folder to zcobol/tests where all current tests reside.
There is an issue with some MULTIPLY operations - most of them are okay though.
These specific cases have been recorded in issue #730 (Incorrect code generated for zCobol MULTIPLY verb) and can be resolved at a later time. That helps to keep this PR from growing too large. Also, we want to reinstate all remaining zCobol regression test programs to reduce the risk of introducing regressions while fixing these multiply issues.


Zaneham (approved):
Approved, this cleans up the old test file and moves everything to /tests

PR #732, opened 2025-12-31
Add build of barcode to BUILD.BAT procedure

abekornelis:
This pull request is part of the effort to re-include all test programs for z390 in the build procedure
Barcode is an optional element, therefore will be assembled only when the build is started as `BUILD.BAT all`
When running the default build (either the BUILD.BAT, the build.sh, or the github-hosted build process) the barcode element is not assmbled.


Zaneham (approved):
Approved

Issue #728, opened 2025-12-20
zCobol fails on multiple programs within a single source file


abekornelis:
See NIST test suite IC222A.CBL through IC401M.CBL
After fixing this issue in zc390.java those test cases need to be activated.

Multiple programs may have conflicting names in their working storage, or they may have conflicting paragraph names, etc.
To cater to that batched assembly is probably required - but z390 does not yet support that.

PR #733, opened 2026-01-01
zCobol should allow nested programs - phase 1

abekornelis:
This is a partial solution to issue #728 
It resolves out-of-sequence error messages when multiple programs are contained in a single source module.
Full resolution requires batch processing to be enabled in z390, but that is currently not supported.


Zaneham (approved):
Approved! tis a good focused change

PR #735, opened 2026-01-03
Fix incorrect label generated for zCobol SUBTRACT with ON SIZE clause

### LINKED ISSUE #734: zCobol subtract with ON SIZE option causes assembly failure
When a cobol program uses a SUBTRACT verb with the ON SIZE option, the generated assembler code (MLC file) causes the assembly to break with the following error:
`AZ390E error 198 (72/1635)14704            JO    PG_SUB_6_ON_SIZE    ERROR IF OVERFLOW`

The cause of the problem is the GEN_SUB macro generating jump instructions to label `PG_SUB_&SUB_LAB._ON_SIZE`, whereas the GEN_ONSIZE_START macro generates a target label using:
`GEN_LABEL SUBTRACT_&SUB_LAB._ON_SIZE,ON_SIZE_ERROR_START`
which results in a label being generated as `PG_SUBTRACT_6_ON_SIZE DS 0H`

Resolution of the problem can be achieved by mending the jump target label in the GEN_SUB macro. There are 5 occurrences of this incorrect label reference.

This issue surfaced from NIST test program NC106A - a trimmed-down version of that program will be used as a test case for validating the fix.

abekornelis:
fixes #734


Zaneham (dismissed):
Approved, GEN_SUB.MAC label fix is straightforward, and the test covers it well

Zaneham (approved):
Rechecked. Looks good :-)

PR #740, opened 2026-01-08
Changes and additions for issue 714

### LINKED ISSUE #714: Problems with sz390.svc_delete()
1. When deleting a previously LOADed module via DELETE EP=<module-name>,
   the storage containing the module is FREEMAINed and the sz390 internal
   CDE parallel arrays are (partially) updated. However, the CDE that was added to
   the chain headed at CVTCDE is not removed from the chain. If the same
   module is subsequently re-LOADed, the LOAD is successful and the sz390
   internal CDE parallel arrays are updated but (1) the CDE on the CVTCDE chain
   is not updated, becoming an "orphan CDE" and (2) a new CDE is not added
   to the chain, resulting in the module not having a CDE on the CVTCDE chain.

2, If a DELETE is done for a module that is not in storage, sz390 issues
       abort_error(88,"delete cde system error")
   while z/OS just sets return code 4 indicating a DELETE failure.

The DELETE processing in sz390.svc_delete() needs to be modified to
1. remove the CDE from the CVTCDE chain and FREEMAIN the CDE
2. zero the CDE parallel array cde_addr[] entry
3. set return code 4 when processing a DELETE request for a module that is not in storage.

jyganci:
fixes #714 
Current sz390.java does not properly manage the CDEs in storage chained from CVTCDE when a module that was loaded is subsequently deleted. When a module is deleted, the SVC DELETE code in sz390.java does not remove the CDE from the CVTCDE chain and release the storage for the CDE. Also, the code in sz390.java should set return code 4 when a DELETE is issued for a module that is not currently loaded. Fixes for these issues are provided in this pull request.


abekornelis (commented):
Please see email with review comments

jyganci:
Haven't seen an email???

abekornelis:
[RunAsmTests.txt](https://github.com/user-attachments/files/24539881/RunAsmTests.txt)
I have reviewed your pull request. These are my comments:

1) well done. Great test code! Thanks for all your efforts.

2) I ran build & tests on both Windows and WSL. No issues were uncovered.

3) The groovy script contains some dependencies. Since groovy may execute tests in arbitrary order,
     I have added code to make the dependencies explicit. See attached. (attachment as .txt file)

4) There is some minimal doco on SVC services. I found doc\reference\program_services.md
     to contain subchapters on LOAD, CDLOAD, DELETE, CDELETE.
    Have you checked the document to upgrade relevant portions of text? Probably we want to
    add some notes about compatible and incompatible aspects of behaviour?
    Could you please have a look at that?

abekornelis (approved):

PR #741, opened 2026-01-09
Changes to support z390 bash directory in path

### LINKED ISSUE #715: Name collisions when putting z390 bash directory in path
A user reported that when putting the z390 bash directory in his path (as suggested by the z390 README.txt file), it was noticed that some of the bash script names collide with existing *nix built-in function names, eg exec and help. This prevents the user from running them with just their names. For example "exec <load-module-name>" executes the *nix built-in function exec. Windows users apparently do not have this problem.

One suggested fix is to use names ending with ".sh" (eg, exec.sh, etc). Another possibility is to prefix all names with "z" (eg, zexec). If either of these is done, it should be done for ALL z390-supplied scripts (asm*, cblc*, etc) that are used to assemble/compile, link, execute source.

The user provided an elegant solution to the problem -- a wrapper script with a name unlikely to collide with anyone else's names. The user provided wrapper script, z3cmd, follows:
```
#!/bin/bash

# Wrapper for "help" and "exec", which are bash
#     builtins conflicting with z390 commands.
#    So, "z3cmd help", or "z3cmd exec [options]".

# Save and discard command name argunent/
cmd390=$1; shift 1

# Read command from directory into current shell
. "$( realpath  ${0%/*} )/$cmd390"
```

jyganci:
Fixes #715.
Some of the z390 bash command names conflict with Linux/MacOS names. If the user puts the z390 bash directory in their path, these name conflicts cause problems. A "wrapper" script is added to avoid the name conflicts.


abekornelis (approved):
Looks good to me.

PR #742, opened 2026-01-09
Pr/replace statement

Zaneham:
Implement COBOL REPLACE statement per FIPS PUB 21-2

Summary

-  Implements REPLACE compiler-directing statement in zc390.java preprocessor
-  Supports `==pseudo-text-1== BY ==pseudo-text-2==` syntax
-   Supports `REPLACE OFF` to deactivate
-   Follows FIPS PUB 21-2 Section XII Chapter 3

  Closes a 16-year TODO (stub created 08/18/2009).

  Many NIST tests use REPLACE and currently fail at the COBOL→assembler stage. This unblocks those tests.


doc/contributing/replace_implementation.md
```diff
@@ -0,0 +1,182 @@
+# zCOBOL REPLACE Statement Implementation
+
+**Author:** Zane Hambly
```
abekornelis, line 3: This is well documented - but can you embed the document in the existing documentation? Without a link to it, few people will be able to find this document.

src/zc390.java
```diff
@@ -185,6 +185,14 @@ h.  Single () wrapped in single quotes.
 	int[]    zc_copy_rep_lst_ix   = null; // RPI 1062 next token on  line index
 	String[] zc_copy_rep_lit1     = null; // RPI 1062 replacing lit1
 	String[] zc_copy_rep_lit2     = null; // RPI 1062 replacing lit2
+    /*
```
abekornelis, line 189: We have an old pre-git habit: the change log needs an entry to document who made what change at which date. And all changed lines should carry the change id (in our case that is the issue number) For the large routine you added, not all lines need to be marked - but mention of the issue number in the routine heading is required instead.
I know it's an old habit - it's almost irrelevant in a git repository. Would you nevertheless please - at least in java code - comply with this habit?

zcobol/mac/REPLACE.MAC
```diff
@@ -20,9 +20,14 @@
...
+.* 08/18/09 RPI stub for pending support
+.* 2025/12 ZH: REPLACE is now implemented as a pre-processor directive
+.*             in zc390.java per FIPS PUB 21-2 Section XII Chapter 3.
+.*             This macro should not be invoked - if you see this error,
+.*             it means the REPLACE statement was not processed correctly
+.*             by the zCOBOL translator.
 .*********************************************************************
          REPLACE
-         MNOTE 8,'REPLACE NOT SUPPORTED YET'
+         MNOTE 4,'REPLACE macro invoked - should be handled by pre-processor'
```
abekornelis: Continuation error! Columns 1-71 are usable for code column 72 is continuation marker when not blank. Please apply correct continuation - or discard the entire macro because:
Since the pre-processor should take care of the replace stuff, we're better off deleting the macro. At least, that's what I think. Would it serve any purpose to keep it around? If the macro is invoked, but it does not exist, then a regular macro-not-found error will be issued by MZ390. What do you think? Does the macro provide any added value?

Additional note: the fact that this issue shows up indicates two things:
1) the changes to zc390 are not effective, this macro should not have been invoked to begin with
2) your testing has not uncovered this issue. How much testing did you do? Should we go over some of the test procedures to give you some guidance

zcobol/tests/TESTREPL.CBL
```diff
@@ -0,0 +1,44 @@
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID.  TESTREPL.
```
abekornelis, line 2: This made me smile! I never knew it would be possible to REPLACE items in open code :-) That's fun - especially in combination with REPLACE OFF. Do we also have REPLACE ON to resume replacement?
Would it be possible to extend this test program with a COPY member that is used twice with and without replace? E.g. a level 01 record structure definition?

zcobol/tests/TESTREPL.CBL
```diff
@@ -0,0 +1,44 @@
...
+      ******************************************************************
+      * Test program for COBOL REPLACE statement implementation.
```
abekornelis, line 4: I like this test program, but it has not yet been embedded in our regression testing infrastructure. Could you please add it to the RUNCBLTESTS bat script, the equivalent bash script, and - above all - to the groovy script. For the bat and bash scripts it is good enough if the test programs compiles and runs, but the groovy test needs to validate that the output is correct as well. There are multiple ways of achieving such validation. Either the program can self-check (that would work in all environments) or the script can apply a diff against a reference file with the expected output.

src/zc390.java
```diff
@@ -185,6 +185,14 @@ h.  Single () wrapped in single quotes.
...
+     * REPLACE statement global variables (2025/12 ZH)
+     * Implements FIPS PUB 21-2 Section XII Chapter 3
```
abekornelis, line 191: Just wondering: I know it's okay in Cobol to continue a quoted literal on the next line. Is it also allowed to span across a line boundary for a numeric literal? A variable? A cobol verb? If the latter is allowed, how does zCobol fare in this respect? Would a line ending in CO followed by a line marked continuation (in column 7) with PY result in the COPY verb being recognized and handled properly? Again - just being curious. Trying to establish the boundary conditions :-)

Zaneham:
Regarding REPLACE ON: According to the COBOL 85 standard (FIPS PUB 21-2, Section XII), only two forms of REPLACE exist:

1. `REPLACE {==pseudo-text-1== BY ==pseudo-text-2==}...` - activates text replacement
2. `REPLACE OFF.` - deactivates all current replacements

There is no `REPLACE ON` in the standard. A new REPLACE statement implicitly replaces any previous REPLACE statement (i.e., it doesn't accumulate). So the pattern is:
- Use `REPLACE ==old== BY ==new==.` to start/change replacement rules
- Use `REPLACE OFF.` to stop all replacements

I've also extended the TESTREPL test to include a third test case with multiple replacement pairs in a single REPLACE statement.

Note: I attempted to add a COPY member test as you suggested, but discovered that the current zCOBOL implementation doesn't support COPY REPLACING with split pseudo-text (`== text ==` with spaces). The REPLACE statement itself works in open code but doesn't affect text from COPY members,  this is consistent with the order of source text manipulation in COBOL (COPY processing happens before REPLACE), though the standard allows REPLACE to affect copied text after COPY text replacement. This could be a future enhancement maybe?

zcobol/tests/TESTREPL.CBL
Zaneham: Ill add this now :-)

doc/contributing/contributing_zcobol.md
```diff
@@ -51,3 +51,9 @@ More details regarding NIST validation can be found in the [zCOBOL user guide](.
 | RC=8 MNOTE support warning        | 19      | 151      | Warning for unsupported items pending implementation |
 | RC=16 At least 1 error message    | 249     | 160      | Error messages from mz390 or az390 macro assembler   |
 
+## Implementation Documentation
+
+The following documents describe recent COBOL statement implementations in detail,
+including normative references to FIPS PUB 21-2 (ANSI X3.23-1985):
```
abekornelis: can you make that a linked reference, please?

doc/contributing/replace_implementation.md
```diff
@@ -0,0 +1,184 @@
...
+
+This implementation conforms to:
+
+> **FIPS PUB 21-2**, *COBOL*, Federal Information Processing Standards Publication,
+> U.S. Department of Commerce, National Institute of Standards and Technology, 1985.
+> Incorporates ANSI X3.23-1985 and ISO 1989-1985.
+
+The following sections of FIPS PUB 21-2 were consulted:
+
+| Section | Title | Page Reference |
```
abekornelis: Can you please align vertical bars for better readability / maintainability in source format?
Same for other tables in this document

zcobol/tests/TESTREPL.CBL
```diff
@@ -0,0 +1,62 @@
...
+       REPLACE ==XXX== BY ==FIRST==
+               ==YYY== BY ==SECOND==.
+
+       01  WS-XXX-VAR       PIC X(10) VALUE 'MULTI 1'.
+       01  WS-YYY-VAR       PIC X(10) VALUE 'MULTI 2'.
+
+       REPLACE OFF.
+
+       PROCEDURE DIVISION.
+       MAIN-PARA.
```
abekornelis, line 47: For better test coverage, could you please pick (or create) a CPZ member and copy that with REPLACING option? Would like to validate that it works equally well on code in the main program text and in copy member source text...

abekornelis (changes requested):
@Zaneham great improvement - I like what you're doing!
What remains are a few minor points. Could you please have a look at those?

doc/contributing/replace_implementation.md
```diff
@@ -0,0 +1,186 @@
...
+This implementation follows the COBOL 85 standard specification. Key compliance points:
+
+- REPLACE operates on source text before compilation
+- REPLACE OFF deactivates all replacements
+- A new REPLACE statement replaces (not extends) prior replacements
+- Pseudo-text must be delimited by `==` markers
+
+## Building
+
+After modifying `zc390.java`:
```
abekornelis: That should not be necessary after we've merged this PR. Could you please remove?

doc/contributing/replace_implementation.md
```diff
@@ -0,0 +1,186 @@
...
+
+```bash
```
abekornelis: 1) Should not be needed after this PR has been merged.
2) We have a pre-built (canned) procedure for doing that. bat\BLDJAR.BAT, or bash/bldjar
3) FYI: there are examples of how to create tabs with BAT/bash instructions on separate tabs.

doc/contributing/replace_implementation.md
```diff
@@ -0,0 +1,186 @@
...
+# Rebuild JAR (include z390.properties)
+copy z390.properties bin\
+jar cvf z390.jar -C bin .
+```
+
+---
+
+## References
+
+1. **FIPS PUB 21-2** (1985). *COBOL*. Federal Information Processing Standards
```
abekornelis, line 157: Would it be possible to link these doc references to where they are currently available on the web? I know such links are subject to wear and tear - but nevertheless ...

zcobol/tests/TESTREPL.CBL
```diff
@@ -0,0 +1,62 @@
...
+           DISPLAY 'TESTREPL - COBOL REPLACE TEST STARTED'.
```
abekornelis, line 48: For better test coverage, could you please pick (or create) a CPZ member and copy that with REPLACING option? Would like to validate that it works equally well on code in the main program text and in copy member source text...

abekornelis (changes requested):
Great progress - well done!
I've inserted a few more detail issues (yes: I'm a real nit-picker)
Could you please have a look at those?

abekornelis (approved):
good enough for now

PR #751, opened 2026-01-19
Assembler should flag MP/DP statements with invalid operand

### LINKED ISSUE #750: Missing error detection for MP assembly
When assembling a MP instruction:
- length of operand 2 must not exceed 7
- length of operand 2 must be less than length of operand1

But these conditions are not detected/flagged during assembly, leading to S0C6 abends at run-time.
The same thing goes for the DP instruction.

Please fix error detection and reporting for MP assembly.

abekornelis:
fixes #750


rt/bat/RUNASMTESTS.BAT
```diff
@@ -29,6 +29,8 @@ call bat\asmlg %z_TraceMode% rt\mlc\IS658    trace sysmac(mac) noloadhigh
 call bat\asmlg %z_TraceMode% rt\mlc\IS658    trace sysmac(mac)                                || goto error
 call bat\asmlg %z_TraceMode% rt\mlc\IS659    trace sysmac(mac) noloadhigh                     || goto error
 call bat\asmlg %z_TraceMode% rt\mlc\IS660    trace sysmac(mac) noloadhigh                     || goto error
+call bat\asm   %z_TraceMode% rt\mlc\IS750    trace sysmac(mac) noloadhigh
+     if %ERRORLEVEL% NEQ 12 goto error
```
jyganci, line 33: I know this can't happen, but if %ERRORLEVEL% were 0, RUNASMTESTS.BAT would prematurely exit with a 0 return code. Are you okay with this behavior? The groovy IS750 test is okay.

abekornelis: It should not happen, but if it would happen the issue should be reported correctly. Thank you for catching this edge case. I'll fix it as soon as I can.

rt/bat/RUNBR14OWE.BAT
```diff
@@ -44,7 +44,8 @@ set z_ReturnCode=0
 goto return
 
 :error
-set z_ReturnCode=%ERRORLEVEL%
+if %ERRORLEVEL% EQU 0 (set z_ReturnCode=6
+               ) else (set z_ReturnCode=%ERRORLEVEL%)
```
jyganci, line 48: It appears to me that all branches to error here are when %ERRORLEVEL% is not 0, so not needed. However, doesn't hurt. Just pointing it out. I'll let you decide if you want to make the change.

abekornelis: Yes, I know. For the batch scripts that follow this particular coding pattern, I have made the same change to each of them, even when technically not needed. If - at some point in the future - I'd need to add a test expecting a non-zero RC, I am bound to repeat the same mistake. So this is intentional, preventive maintenance.

jyganci (approved):

PR #757, opened 2026-01-29
Fix Multiply issues in zCobol

### LINKED ISSUE #730: Incorrect code generated for zCobol MULTIPLY verb

abekornelis:
Fixes #730

To fix GEN_MPY issuing invalid MP instructions, length checks have been added for source and target field. Fields are lengthened when needed. When a single MP cannot do the job (shortest operand 8 bytes long or more) we generate a matrix multiplication.

Disabled tests in TEXTCMP1 and TESTCMP4 have been re-enabled.
As a side effect NIST program NC220M that failed S0C6 now runs to completion with RC=0.

The results for running the NIST test suite with the changes in this PR are covered by PR#25 in that repo.


zcobol/tests/abe.CBL
Zaneham, line 1: Is this a personal test file? Just wondering if it should be committed or excluded

Zaneham (approved):
Approved, thanks for removing the temp test files.

PR #758, opened 2026-01-31
Import CICS contribution document

abekornelis:
The document existed in PDF format, but was never imported into the repository.


adelosa (commented):
Have you tested this process for the latest version. Some of the old docs reference commands that have long disappeared? Also, looks like a recent merge added an unrelated AI generated doc to contributing section - cobol replace? I'm not a fan of putting people's names in the docs. We can have an acknowledgment page for that but docs themselves should be without attribution. Hope that makes sense.

doc/contributing/contribute_cics.md
```diff
@@ -0,0 +1,113 @@
...
+
+This error is caused by a structural mismatch between `SEQO0001.TXT` and the comparator files SEQCnnnn.
+The output file `SEQCOMP.TXT` will contain an error message and a full dump of both files.
+
+Before reporting these as errors, please check the following:
+1. Are all the SEQI and SEQC files in sequence ?
+2. Has a simulation run been done using the SEQI files (`SEQ_TERM=YES`) ?
+3. Have any of the transactions changed since the last run of `Z390SEQ` for those transactions ?
+
+## Trademarks
```
adelosa: I have been removing credits from the web docs. My approach to this is that the docs belong to the project. If you want attribution or somewhere we can credit contributors, then lets have a single page for that rather than spreading it through the docs.
Same deal for trademarks - You only need one spot for this and no need for every page.

abekornelis: I have made additional changes to meet your requirement. Should be okay now, although I may have missed navigation updates for the generated pages.

adelosa (approved):
This has BAT file changes which I can't test as I don't use windows. You will need to get someone who uses Windows to review this. The docs look fine. I am approving this so you can decide if you want to get someone to test the BAT file changes.

PR #761, opened 2026-02-13
Fix pz390 DP handling for S0CB

### LINKED ISSUE #760: The pz390.java code issues S0CA when decimal divide (DP) produces quotient too large to fit
The pz390.java processing of the DP instruction incorrectly produces S0CA (decimal overflow) when a division results in a quotient that is too large to fit in the first operand field. Principles of Operation states that "Overflow cannot occur. If the divisor is zero or the quotient is too large to be represented by the number of digits specified, a decimal-divide exception is recognized". S0CB (decimal divide) is produced when attempting to divide by zero but not when a quotient is too large to fit. Sample code that illustrates the issue.
```
TDP      CSECT
         USING *,15
         DP    P1,P2       z390 S0CA; should be S0CB
         DS    0D
         DC    C'AAAAAAAA'
         DC    PL2'999'
         DC    PL1'1'
         DS    0F
P1       DC    PL2'999'
P2       DC    PL1'1'
         DS    0F
*
* Dump output snippet
* ... ABEND PSW=07850600 80008000 FD10F014F016 DP    ABEND S0CA
*
* 00008000 *FD10F014 F0160000 C1C1C1C1 C1C1C1C1* *..0.0...AAAAAAAA*
* 00008010 *999C1C00 9C9C1C00 00000000 00000000* *r...............*
*    before ----==   ----== after
*           P1  P2   P1  P2
*
         END
```

jyganci:
Fixes #760.

Also added tests for DP S0C6. Fixes: pz390.java. New tests: tests/TESTINS2.MLC.

### DIFF
--- a/src/pz390.java
+++ b/src/pz390.java
@@ -401,6 +401,8 @@ public class pz390 {
      * 2022-03-12 DSH fixed overflow for cc3 MSC, MSGC, MSGRKC, MSRKC
      * 2023-03-29 Issue 434 ED instruction handles alternate signs incorrectly
      * 2025-10-20 AFK       Add javadoc comments
+     * 2026-02-06 Issue 760 Divide Decimal (DP) incorrectly issues S0CA when quotient
+     *                      length too long; should issue S0CB
 	 *********************************************************
 	 * Global variables              (last RPI)
 	 ********************************************************/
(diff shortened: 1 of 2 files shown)


abekornelis (approved):
Looks good to me.

PR #762, opened 2026-02-15
Define new version to create release

abekornelis:
I've created release for V1.8.4.2
The choice is a bit arbitrary - we could just as well  define 1.8.5


jyganci (approved):
Okay as is.

PR #763, opened 2026-02-15
Fix javadoc comments in test_screen and test_unicode

abekornelis:
no issue apples


BobMissing (approved):
Rearranging of comments in the two source modules.  No logic changes.

PR #764, opened 2026-02-19
Import zVSAM design document

abekornelis:
This is the initial version, plus maintenance applied until Melvyn Maltz took over.


Zaneham (approved):
Approved as this is just adding already existing documentation into the repository.

PR #774, opened 2026-03-15
Converted Melvyn's 2.4 version of zVSAM V2 design document to markdown

abekornelis:
No issue applies to this change.
I converted the PDF file that Melvyn created: zVSAM V2 Design and Logic V2_4.pdf
The resulting markdown files and the associated diagrams are in this PR.
The diagrams are present as xml files - as saved from draw.io
and as .pdf files that are linked in with the markdown files.


Zaneham (approved):
Approved thanks for getting this done

PR #776, opened 2026-03-19
Make the two design document versions comparable

abekornelis:
No issue applies to this change
I'm reordering text sections and applying small changes to make the two versions of the design document comparable.

(the next step after completing this PR should be to evaluate the two versions by comparing them and then create a new version that can serve as a basis for actual development)


abekornelis:
Closing this PR - created new PR #782 to replace this one.

PR #777, opened 2026-03-22
Fix zCobol test cases with muliply packed issue

### LINKED ISSUE #759: zCobol MULTIPLY verb may generate code for S0C7
For MP the architecture prescribes a specification exception (S0C6) when either:
- the length of operand 2 exceeds 8 bytes, or
- the length of operand not less than length of operand 1

The macro GEN_MP takes this into account.

However the architecture also prescribes a decimal data exception (S0C7) when:
- the operand 1 does not contain as many bytes of zeros as the number of bytes in operand 2.

The macro GEN_MP does not take this into account when selecting the multiplication strategy.

Please extend GEN_MP to cater to the leading zeroes requirement for MP.

abekornelis:
fixes #759 
For now I have changed the test cases - they were multiplying numbers that are simply too large for MP to handle.
I.e. MP requires operand1 to have as many leading bytes of zeroes (2 packed zeroes each) as the entire length of operand2. The operand2 in this case was assembled as 4 bytes, but operand 1 did not have 8 leading zeroes.


Zaneham (approved):
Approved. Happy for this to merge :-)

PR #779, opened 2026-03-25
Flag invalid register operands for KM, KMA, and KMC

### LINKED ISSUE #704: Assembly of KM/KMA/KMC fails to flag invalid register operands
Reported by Dan Greiner by email dd 2025-10-23 06:45

I noticed a minor inconsistency in how the z390 assembler detects incorrect register usage.  

For instructions such as CLCL and MVCL, the assembler flags an error if either register is odd.
However for some newer instructions that have similar constraints (like KM, KMA, KMC), the assembler blithely ignores incorrect register specifications. 

For the later cases, detection of incorrect registers is further complicated by disallowing register 0, and in the case of KMA, R3 cannot be the same register as R1 or R2.

Zaneham:
Assembler now flags error 56 for invalid register usage on KM, KMC, and KMA instructions (even required, R0 disallowed, KMA R3 must differ from R1/R2).

Fixes #704


abekornelis:
@Zaneham - looks good. Just a few remarks on your testing logic. The test program is validated for RC=12. This does not validate individual error messages, and ignores any issues (should they ever arise) with the happy-path tests that you also included in the test program.
I suggest:
- assemble with notiming option
- create copy of the .err file as a .tf1 file
- add a comparison step to each test (bash, bat, groovy) to ensure the .err file remains unchanged
- file comparison steps already exist, you can copy the basic logic from e.g. TESTFPC2

abekornelis:
@Zaneham A second minor issue: in the java code we (try to) consistently add change markers. You've done so on several lines, but not all. Could you please use the delta overview in this PR to find all changed locations and make sure all modified lines are marked with a #704 in the comments?
Also, the change log in the program's heading needs an entry to log the change event.

abekornelis (changes requested):
Please see the two comments I left.
Nothing big - please re-request review when done?

Zaneham:
Hi @abekornelis, addressed both review comments:

**Comment 1 (test validation):**
- Assemble with `notiming` so the .ERR file is deterministic
- Captured the expected output as `rt/mlc/IS704.TF1` (35 lines, 16 errors)
- bash/bat/groovy runners now diff `.ERR` against `.TF1` after the RC=12 check, so any drift in the happy-path tests OR the error message text will fail the build
- Pattern modelled on TESTFPC2

**Comment 2 (#704 markers + changelog):**
- Added `// #704` to every modified line in `src/az390.java` (including braces in the new KM/KMC/KMA blocks and all lines of the new `get_hex_reg_even_nz()` helper), right-aligned per the #730/#750 convention
- Added a changelog entry: `2026-03-25 ZH  #704 Flag invalid register operands for KM, KMA, KMC`

Re-requesting review. Thanks!

abekornelis (approved):
Passes all tests.

PR #780, opened 2026-03-28
Fix Getmain Issue with subpool in length register

### LINKED ISSUE #773: Getmain issue with subpool nr
Reported by Martin Ward, email dd 2026-03-12 18:01

I am testing a mainframe module using z390 and GETMAIN was abending
with S80A. It turned out to be due to the requested size being passed
in a register via LV=(2) where the high byte of the register was
set to specify the GETMAIN subpool on the mainframe.
So it was requesting a size of 0x07000050 instead of 0x00000050

Could the z390 GETMAIN ignore this first byte, or is it needed
for really large memory allocation requests?

abekornelis:
fixes #773


martin-ward (approved):
Looks good

PR #783, opened 2026-04-26
Add TESTVSM4 to dfhall procedure

abekornelis:
This PR carries two changes:
1) TESTVSM4 was not part of the dfhall / DFHALL.BAT scripts. I have added a step to assemble TESTVSM4
2) The UNREF steps take a lot of time, and they're a development tool only. I changed DFHALL.BAT to skip UNREF steps by default. This change has now been carried over to the dfhall bash script: by default it will skip the UNREF steps. Only when invoked with unref argument will it include the UNREF steps.


jyganci (approved):
Looks good. I did not know about bash ${1,,} converting $1 to lowercase and the use of a leading colon on a statement line.

PR #793, opened 2026-05-04
Fix remaining javadoc issues

abekornelis:
No issue related to this fix for last two remaining javadoc warnings.
two issues kept javadoc from generating doc for classes ArchLevel and sz390.
Those issues have been fixed - the javadoc now generated without warnings.


adelosa (dismissed):
I would personally not add the comment about the prior placement. It works now, so why mention it. There is a record in source control if anyone was to trace the file history. Approving for you to make the call.

abekornelis:
Hi Anthony, thanks for spotting that comment - I should have noticed myself. I've taken it out, it no longer makes any sense.
Could you please re-approve?

jyganci (approved):

PR #794, opened 2026-05-04
Add java to index.md page

abekornelis:
Making explicit what was implicit - z390 is java-based and open-source


adelosa (changes requested):
To me, GPLv2 licensed and open-source are related concepts and should be together. I did target the content of this page to users of the app rather than people who want to help develop it. Clearly the Java part is really only interesting to devs as users just install prereqs (A Java runtime). So it was a deliberate exclusion, but if you want to have something on the homepage, go for it. Maybe just say Works with Java 8+.

abekornelis:
Hi Anthony, That's okay - I've changed the wording. I hope you like it better this way.

jyganci (approved):

PR #796, opened 2026-05-08
Add doco on zCobol trace handling

### LINKED ISSUE #795: During abend incorrect PSW is reported
While testing YouzPi2.CBL a S0C7 abend occurred which was reported at offsetX'3D2'
When running the exact same test with TRACE option added, the .TRE file shows reports abend S0C7 at offset X'426'

Turns out the trace is incorrect - it reports an offset that is in the middle of an instruction according to the .PRN file.

The cause is that several zCobol macros expand differently in TRACE mode.
I've added text to the documentation to clarify the issue.
Conclusion: the PSW is only seemingly incorrect - due to the abend location changing with absence or presence of a TRACE option. How confusing!

abekornelis:
fixes #795

### DIFF
--- a/doc/contributing/contributing_zcobol.md
+++ b/doc/contributing/contributing_zcobol.md
@@ -31,6 +31,22 @@ written in zCOBOL which verify that zCOBOL statements
 produce the expected results. And finally there is a need to develop documentation
 on the zCOBOL project as it evolves.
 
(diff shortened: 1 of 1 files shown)


Zaneham (approved):
Approved, thanks for the addition.

PR #803, opened 2026-05-14
Revive guam demo and test programs

abekornelis:


Zaneham (approved):
Passes on my windows machine, looking good!

PR #805, opened 2026-05-15
Create overview of Java V8 issues

abekornelis:
When compiling our java code with Open JDK 21.0.6 we have some issues. These are detailed in this new document.
Please mind: the bldjar.bat script has been modified. The bash script bldjar will have to be modified as well, sometime later.


abekornelis:
I have created Issues #809 #810 #811 to cover the problems. This PR no longer needed.

PR #806, opened 2026-05-15
Create initial linter for z390 java code

### LINKED ISSUE #807: Add minimal linter stub to build process
Please add a stub for a minimal linter as part of the local build process (bat\BLDJAR.BAT and bash/bldjar) to prove the feasibility of including a linter step in the build process.

abekornelis:
fixes #807 
For now: only 1 check (for unused imports) and only in the local build on windows.
Making a comparable change to the bash/bldjar process is okay.
But we must make sure the build process on github remains unaffected until we resolve the errors reported ...


bat/BLDJAR.BAT
```diff
@@ -32,6 +32,16 @@ shift /1
...
 
+rem download linter jar
+if not exist "build\checkstyle"                    mkdir build\checkstyle
+if not exist "build\checkstyle\checkstyle-all.jar" powershell -NoProfile -ExecutionPolicy Bypass -File "bat\GetCheckStyle.ps1"
+if not exist "build\checkstyle\checkstyle-all.jar" goto return || rem download must have failed - cannot continue
+java -jar build\checkstyle\checkstyle-all.jar --version
+
+rem run the linter
+java -jar build\checkstyle\checkstyle-all.jar -c src\checkstyle.xml src
+if %ERRORLEVEL% NEQ 0 goto return
```
Zaneham, line 43: This runs over the whole `src` tree before javac, so merging this on its own stops the Windows build dead. I ran checkstyle 13.9.0 with your config over `src` at this head and got 108,938 violations across 20 files, mostly pz390, mz390, az390 and sz390. Even once 895-898 are in, about 93,000 are left.

So either this goes last, or the gate wants to be warn-only for now. A suppression file listing the not-yet-clean modules would let it be a real gate from day one and shrink as you work through them.

bat/GetCheckStyle.PS1
```diff
@@ -0,0 +1,39 @@
+# Suppress Invoke-WebRequest's built-in progress bar
+$ProgressPreference = 'SilentlyContinue'
+
+# Request latest release information
+$apiUrl = 'https://api.github.com/repos/checkstyle/checkstyle/releases/latest'
```
Zaneham, line 5: `releases/latest` means whoever builds gets whatever happened to be current when they first ran it, and a new checkstyle release can fail the build with no code change. Latest is already 13.9.0 against the 13.4.2 in your comment above. You write `version.txt` on line 30 but nothing ever reads it, so pinning a version and comparing against that file would fix both at once.

Zaneham (commented):
Hey Abe, I've had a proper look at this one. The ruleset itself looks good. One thing that isn't inline anywhere because it's about what's missing rather than what's here, the linter only runs from `bat/BLDJAR.BAT`. CI builds on ubuntu-latest through `bash/blddist` and `bash/bldjar`, so nothing here ever fires on a PR. That means the cleanup in 895-898 can regress without anyone noticing, and contributors on Linux never see it at all. Should `bash/bldjar` get the same block, or would you rather it was its own CI job?

bat/GetCheckStyle.PS1
```diff
@@ -0,0 +1,39 @@
...
+    $tag | Set-Content -Path $versionFile -NoNewline
+    Write-Progress -Activity "Download" -Completed
+    Write-Host "Download finished: $tag"
+    exit 0
+}
+catch {
+    Write-Progress -Activity "Download" -Completed
+    Write-Host "Download failed: $($_.Exception.Message)" -ForegroundColor Red
+    exit 1
+}
```
Zaneham, line 39: No trailing newline on this one.

abekornelis:
Hi @Zaneham, You are correct - this is still a work in progress. It started out as a minimal thingy, just to prove feasibility. Then got out of control more or less. So, yes, the plan is indeed to fix all the java programs before merging this change. And an update to the bash build script definitely will have to be part of that merge. That's the whole point of having a linter to begin with.

Zaneham (dismissed):
Looks good to me on my end :-)

bash/blddist
```diff
@@ -56,12 +56,6 @@ cp -R ${ROOT_DIR}/build/zstrmac/* mac/spm
 mkdir -p javadoc
 cp -R ${ROOT_DIR}/build/javadoc/* javadoc
 
-# remove components not required for distribution
```
jyganci, line 59: Why are you removing the removal of components not required for distribution? Is it not true that the src directory is not part of the distribution? Yes, a user can separately get the source, but not as part of the distribution. Correct?

bash/bldjar
```diff
@@ -5,6 +5,18 @@
...
+if [ ! -f ../build/checkstyle/checkstyle-all.jar ]; then
+    ./getcheckstyle
+fi
+if [ ! -f ../build/checkstyle/checkstyle-all.jar ]; then
+    echo "bldjar: linter download failed - cannot continue"
+    exit 1
+fi
+java -jar ../build/checkstyle/checkstyle-all.jar --version
+java -jar ../build/checkstyle/checkstyle-all.jar -c ../src/checkstyle.xml ../src
+
```
jyganci: I personally have no desire to run this portion every time I run bldjar. I definitely do not want to have this script silently download and execute software to my PC.

Since I use a modified bldjar script on my PC, this will not affect me.

bash/getcheckstyle
```diff
@@ -0,0 +1,24 @@
...
+
+# Follow /releases/latest to the real tag URL, then take the last path part.
+latest_url=$(curl -fsSL -o /dev/null -w '%{url_effective}' \
+    https://github.com/checkstyle/checkstyle/releases/latest)
+tag=$(printf '%s\n' "${latest_url}" | sed 's|.*/||')
+jar_url="https://github.com/checkstyle/checkstyle/releases/download/${tag}/${tag}-all.jar"
+
+curl -fL --retry 3 -o "${out_file}" "${jar_url}"
+printf '%s' "${tag}" > "${version_file}"
+echo "Download finished: ${tag}"
```
jyganci: Missing newline at end of source.

bat/BLDJAR.BAT
```diff
@@ -32,6 +32,16 @@ shift /1
...
+
```
jyganci, line 53: Agree with Zane's comments. Also applies to the bash script.

src/checkstyle.xml
```diff
@@ -0,0 +1,112 @@
...
+  <!-- Require three empty lines after the last import -->
+  <module name="RegexpMultiline">
+    <property name="matchAcrossLines" value="true"/>
+    <property name="format"
+      value="(?m)^import .+;(?: *//[^\r\n]*)?\r?\n(?:\r?\n){0,2}/\*"/>
+    <property name="message"
+      value="Need 3 empty lines after the last import."/>
+    <property name="fileExtensions" value="java"/>
+  </module>
+</module>
```
jyganci, line 114: Don't really know XML, so cannot make any meaningful comment. However, the src directory is for Java, with 2 exceptions -- Z390.IVP and Z390.MAN (and Z390.IVP is obsolete, is it not?). Does the XML file really belong here?

src/gz390.java
```diff
@@ -671,7 +671,7 @@ private int set_main_mode(String[] args) {
         if (!perm_file_log) {
             perm_file_write = false;
         }
-        /******************************************
+        /* *****************************************
```
jyganci, line 674: Okay, but why add the space?

src/az390.java
```diff
@@ -573,7 +573,7 @@ public  class  az390 implements Runnable {
                       ((long)(-1) >>> 17),
                       ((long)(-1) >>>  9),
                       ((long)(-1) >>>  1),
-                    };
+        };
```
jyganci, line 576: I prefer the original line, but okay. We've already discussed this with my z390.java changes.

abekornelis: Basically, I agree. But am hesitant to change the linter rules at this stage. However, changing the linter rules later is fine with me, by all means - it's a learning/discovery process.
There is an alternative that the current linter does allow: to put the terminating } at the end of the previous line, rather than on  a line by itself. I can make that change, if you want me to.

jyganci: Leave it as is.

bash/blddist
abekornelis: To my knowledge, the GNU license requires us to distribute the means to recreate the distribution from source. So we can leave out zCics (and it sources) but we are required to include all that is needed for the recipient to continue development from their copy, even if github goes belly-up for whatever reason.

bash/bldjar
abekornelis: I agree to remove automatic execution of the update script. I'll make that a separate script one can run when desired.

bash/getcheckstyle
abekornelis: fixed.

bat/BLDJAR.BAT
abekornelis: Together we have fixed all the 100K+ errors. At least in that respect we should be okay now.

bat/BLDJAR.BAT
abekornelis: I have made sure all source code passes the linter tests. We should be okay now.

src/checkstyle.xml
abekornelis: Yes it does - it really is a source file specifying to the linter tool what checks to apply.

src/gz390.java
abekornelis: Without the space the comment block is seen - by the linter - as a javadoc comment and associated rules are applied. I do not remember the 'issue' it reported, but this fixed it.

bash/blddist
jyganci: I agree with your response. You did not answer one of my questions. Will the src directory now be part of the release or will the user have to download it separately in order to build z390. Just wondering.

src/gz390.java
jyganci: Ah, I should've figured that out for myself. For Java I use
```
/*
 **********************************************
 * comment
 **********************************************
 */
```
if I want a comment box. Similar to what I do for an assembly language box comment.

Zaneham (approved):
Thanks for making the changes :-) I'm still happy with this.

PR #815, opened 2026-05-18
Add mfacc programs to test suite

abekornelis:
There is no issue associated with this PR


Zaneham (approved):
Approved, Ran tests on my windows machine and had a skim of the generated code. Used Ollama as a second 'pair of eyes' (even though it's an LLM) due to the amount of generated code.  All 65 mfacc programs assemble, link and run RC=0 under the existing toolchain.

PR #835, opened 2026-05-29
Add IVP scripts and integrate with build procedure

### LINKED ISSUE #833: Revive the IVP
The java code for IVP still exists.
Although not very important, let's dig up the IVP.BAT and bash scripts so users can run the IVP and report OS level and z390 level when reporting an issue. It should be dead easy for the user to create an accurate and dependable version report.

abekornelis:
Fixes #833


Zaneham (approved):
All good here.  Only thing I can think of is `bash/ivp` has no set -e but it captures and re-exits $?, so failures still propagate. It should be fine

PR #839, opened 2026-06-08
zCobol GO TO DEPENDING: scan for DEPENDING (any target count), allow optional ON

### LINKED ISSUE #891: zCobol GO TO DEPENDING not handling optional ON or a subscripted index
Two defects in `GO TO ... DEPENDING`, both present in NIST NC1024.

**1. `ON` is treated as required, and the target scan uses a fixed offset.**

`ON` is optional in the standard. `GO.MAC` also looked for `DEPENDING` at a fixed offset (`&PARM_IX LT &NP-2`) rather than scanning for the keyword, so forms with few targets were missed even with `ON` present.

```cobol
       MOVE 2 TO IX1
       GO TO Q1 Q2 DEPENDING IX1
```

```
MZ390E MNOTE 8,'GO TO DEPENDING CLAUSE NOT FOUND'
MZ390E MNOTE 8,'GO TO DEPENDING VAR NOT FOUND - '
```

NC1024 uses both: `DEPENDING GO-TO-DEPEND` with no `ON` at GO--TEST-F2-3, and a single-target form at GO--TEST-F2-6.

**2. A subscripted index abends at runtime.**

The index address is in a register, but the SS-format `ZAP`/`PACK` were given `off(reg)`, which an SS operand reads as `off(length)` rather than `off(len,reg)`. It assembles cleanly and then abends:

```cobol
       MOVE 2 TO ITEM(3)
       GO TO PD PE PF DEPENDING ON ITEM(3)
```

```
EZ390E error  11 ABEND PSW=07850600 800FFE26 F87090A890B0 ZAP   ABEND S0C7
```

NC1024 GO--TEST-F2-5 uses `DEPENDING ON GO-SCRIPT (7)`. NC1074 NUM-TEST-2 also needs numeric paragraph names to select by position rather than by name.

The NIST suite cannot currently demonstrate any of this. Of the 27 programs adopted in the harness only IC227A1 uses `GO TO DEPENDING`, and it stops at `FD OPTION NOT SUPPORTED - EXTERNAL` before reaching the statement. NC1024 and NC1074 are not adopted. Coverage has been added to `zcobol/tests/TESTGO1.CBL` instead, which already runs from all three script sets.

Zaneham:
fixes #891


Zaneham:
NC102A now assembles but triggers a SOC7 error on my end. If anyone is willing to run the NIST tests on this and let me know what happens please do. It could just be an issue with my testing harness/environment so going to check.

abekornelis (changes requested):
The new test program has not yet been embedded in the bat/bash/groovy scripts.
Also, what problem is being fixed? Can you please create an issue to describe the problem?

I ran the NIST tests anyway. Most GO TO DEPENDING ON errors still persist. See e.g. IC223A

PR #842, opened 2026-06-20
Move design documents to design folder and fix navigation

abekornelis:
Moving zCobol design documents from the user_guide subfolder to the design subfolder.


doc/design/zCOBOL/.pages
Zaneham, line 1: doc/design/.pages nav lists - zCOBOL, but the folder actually is doc/design/zCobol/. mkdocs/Linux are case-sensitive, so that nav entry won't fully work.

Zaneham: This has bitten me a couple times

Zaneham (changes requested):

doc/design/zCOBOL/.pages
abekornelis: @Zaneham As per our standards I have renamed the zCobol subdir to zCOBOL
Could you please re-review?

Zaneham (dismissed):

Zaneham (approved):

PR #845, opened 2026-06-23
Interfaces

abekornelis:


abekornelis:
Other changes in the doc structure made this change tricky to carry on. We'll carry the changes over to a fresh branch.

PR #847, opened 2026-06-25
Fix shortcomings of bash/ivp script

### LINKED ISSUE #857: Inconsistent behaviour of IVP procedure
For me, on Windows, the IVP returns when run from the root:
```
IVP OS   version = Windows 11
IVP J2SE version = 21.0.6
IVP z390 version = 1.8.4.3
IVP exiting with RC = 0
```

Yet for John, on Linux, it returns when run from the root:
```
$ pwd
/home/john/z390/work/z390
$ bash/ivp
IVP OS   version = Linux
IVP J2SE version = 21.0.7
IVP z390 version = v1.8.4.3jg  I added the trailing jg
IVP z390 base version file Z390.IVP not found
IVP see errors above
IVP exiting with RC = 8
```

And for John, on Linux, it returns when run from the src directory:
```
$ pwd
/home/john/z390/work/z390/src
$ ../bash/ivp
IVP OS   version = Linux
IVP J2SE version = 21.0.7
IVP z390 version = v1.8.4.3jg
IVP z390 base install not equal z390.jar PTF version
IVP RT optional  zip not installed
IVP MVS optional  zip not installed
IVP see warnings above
IVP exiting with RC = 4
```

Please make IVP behaviour consistent, and remove references to optional components - those have been merged into the standard distribution long ago.

abekornelis:
there is no issue describing this change
Fixes #857 which covered initial review results.


bash/ivp
```diff
@@ -1,17 +1,33 @@
...
+
+if [ $trace -eq 1 ]; then set -x; fi
+
 # get full directory path to z390.jar; assumes that this bash script is in the z390 subdirectory
-jardir=$(dirname $0)
-jardir=$(cd $jardir && pwd)
-jardir=$(dirname $jardir)
+jardir=$(dirname "$0")
+jardir=$(cd "$jardir" && pwd)
+jardir=$(dirname "$jardir")
```
jyganci: I assume you are putting double quotes around $0 in case $0 contains spaces or some other special characters. Correct? Why are you doing this now? There are MANY other bash scripts that issue jardir=$(dirname $0). Should all of them also be changed? The bash scripts are used within the confines of the z390 directory. No directory names, paths, etc contain spaces.

abekornelis: @jyganci I do not really understand bash scripts. I involved AI to help me make the changes. I followed the suggestions that seemed relevant. I can take out the quotes if you want to. I have no specific affinity with them. I would rather follow your advice than some AI engine ...

jyganci: That was my guess as to why you did it. (I did a google search; the AI summary was as you saw.) Look in bash/az390 for how it is done everywhere else in z390. Yes, without the surrounding quotes there is a possiblity of an error if the path name has spaces in it. However, there are no such path names in the z390 directory. If it were me, I would do like what is done in bash/az390. I'll give you a chance to either do that or tell me you will leave it as is. In either case, after looking at the PR to verify,  I will then approve it.

jyganci (approved):

PR #858, opened 2026-07-03
Revive rt scripts for Opcode checks

abekornelis:
There is no issue describing this change.
What I did:
- Fix a few minor flaws in the `OPCD_*` programs, their options (`*.OPT`) and their validations (`*.RT`)
- Add as testcases to the `RUNSCRIPTS.BAT`, `runscripts` (bash), and `RunScripts.groovy` test scripts


Zaneham (approved):
Tested on windows

PR #859, opened 2026-07-03
Suppress build/test on github for doc-only changes

abekornelis:
There is no issue to cover this change.

The purpose of this change is to suppress the full build/test run on the github server when a PR contains only doc changes.
I have modified the build.yml file to trigger the build/test process only when:
- any file has changed
- with the exception of .md and .txt files (case insensitive)
- yet treat the version.txt file as a normal file to trigger the build.

As a result the build/test will be suppressed when there are only changed .txt and/or .md files in the pull request, while a change to the version.txt file will still trigger the build/test process.


Zaneham (dismissed):

Zaneham:
Apologies, wires crossed there. Have reverted the approval :-)

adelosa:
Not sure what problem you are trying to fix, but whatever.
My personal approach would be to add the following in the main build instead.
This will skip builds if only those paths are modified which I think is what you are after.
There is no reason to be updating docs outside of these folders.
```
on:
  push:
    paths-ignore:
      - 'doc/**'
      - 'doc_overrides/**'
      - 'mkdocs.yml'
  pull_request:
    paths-ignore:
      - 'docs/**'
      - 'doc_overrides/**'
      - 'mkdocs.yml'
```

adelosa (commented):
See comment

abekornelis:
@adelosa That is a very valid alternative approach. I'll give it a try as it's a cleaner approach.
The reason for doing this is that the full regression test takes ca. ten minutes - which okay, but a waste of time and resources when we have a doc-only change. And we're having lots of doc-only changes at the moment, because we're working to upgrade the zVSAM V2 design docs and doing so in small increments. Thanks for the suggestion. I'll update the PR in a bit.

jyganci (approved):

PR #864, opened 2026-07-05
Fixes for execution of TRT, TRTR, ED, EDMK running in Amode 64

### LINKED ISSUE #596: TRT in Amode64 does not clear high word of R1
Reported by Bob Schulien, email dd 2025-01-16, 10:08

The TRT instruction does not clear the high half of R1 if the CPU is in AMODE 64.
      LGHI   R1,-1
     TRT     FIELD,TRT_TABLE
     IF (NZ)
          R1 high half is still high-values/not a 64 bit address.
    ENDIF
Workaround is to do a LGHI R1,0 before the TRT.

jyganci:
Fixes #596. Issue #596 reported that the TRT instruction did not correctly set 64-bit register 1 (GR1) when issued in 64-bit addressing mode. Research unconvered two more instructions -- TRTR and EDMK -- that have the same problem. Testing a preliminary fix uncovered additional problems with the ED and EDMK instructions. Finally, running the Groovy regression test for the mfacc subdirectory uncovered a bug in two of the test programs -- P9MM1.MLC and P9MM2.MLC -- where R1 was not set correctly prior to an EDMK instruction. Previous pz390.java EDMK code did not properly handle the invalid R1 value, resulting in the two programs appearing to run correctly.

All of these issues/problems are being fixed by this pull request.

### DIFF
--- a/rt/mlc/IS596.MLC
+++ b/rt/mlc/IS596.MLC
@@ -1,0 +1,4364 @@
+***********************************************************************
+* Regression test for issue #596:
+*     TRT in 64-bit addressing mode does not properly set
+*     64-bit general register 1 when match found.
+*
+* Two additional instructions, TRTR and EDMK, are also impacted.
+*
+* TRTR also incorrectly sets bit 32 of register 1 when executed
+* in 31-bit addressing mode.
+*
+* Testing uncovered additional problems with the ED and EDMK
+* instructions:
+*   1.  ED and EDMK should terminate with a data exception
+*       if the left digit in a byte of the packed decimal source
+*       is greater than 9 (B'1010' -- B'1111'). In this case, the
+*       edit pattern and condition code should be unchanged for
+*       both ED and EDMK; for EDMK, GR1 should be unchanged. An
+*       abnormal temination with system code S0C7 should result.
+*   2.  ED and EDMK incorrectly set the condition code for an
+*       unsigned packed decimal source field such as X'1234'.
+*   3.  EDMK incorrectly set GR1 in cases where the register should
+*       remain unchanged.
+*   4.  Running the Groovy regression tests uncovered a bug in two
+*       mfacc programs -- P9MM1.MLC and P9MM2.MLC. They both
+*       incorrectly set R1 before an EDMK. The current pz390.java
+*       changes fixed a pz390.java bug that allowed the two mfacc
+*       programs to (apparently) work correctly. These two mfacc
+*       programs are being modified to correctly set R1 before an
+*       EDMK.
+*       
+* These problems are also being fixed by #596.
+*
+* The tests performed here verify that
+*   1. TRT and TRTR correctly set GR1, GR2 and the condition code
+*      when running in all three addressing modes.
+*   2. ED and EDMK correctly edit the source and correctly set
+*      the condition code for both signed and unsigned packed
+*      decimal values. For ED, the tests run in addressing mode 31.
+*      For EDMK, the tests are run in all three addressing modes
+*      in order to verify that EDMK correctly sets 64-bit GR1.
+*   3. ED and EDMK correctly detect and handle invalid packed
+*      decimal source, terminating with a data exception and
+*      leaving the original edit pattern and condition code
+*      unchanged. In addition, EDMK leaves 64-bit GR1 unchanged.
+*   4. The two mfacc programs are verified to run correctly when
+*      the Groovy regression test for mfacc is run.
+*
+* Registers on entry:
+*     15  entry point
+*     14  return point
+*     13  standard 18 word save area
+*
+* Registers on exit:
+*     0--14  as at entry
+*     15     return code
+*                0  all tests were successful
+*                8  at least one test failed
+*
+* Output:
+*     Informatory WTOs are issued for each test
+*     Error WTOs are issued when an error occurs
+*
+* Restrictions: The test program must reside in 24-bit storage and
+*               z390 option "noinit" must be used when assembling
+*               and linking the program.
+*
+*               The ED and EDMK tests have additional restrictions.
+*               See the comments for the ED tests.
+*
+*               Maximum total number of tests is 99,999.
+**********************************************************************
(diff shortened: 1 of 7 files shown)


rt/mlc/IS596.MLC
```diff
@@ -0,0 +1,4364 @@
...
+*
+***********************************************************************
+*        Strings used in the TRT and TRTR tests
+***********************************************************************
+         DS    0D
+StrNone  DC    C'abcd'
+StrMid   DC    C'a  d'
+StrEnd   DC    C'abc '
+StrBeg   DC    C' bcd'
+         DC    16X'00'         Needed to avoid S0C5 due to pz390 bug
```
abekornelis: Could you please create an issue to describe this bug?

rt/mlc/IS596.MLC
```diff
@@ -0,0 +1,4364 @@
...
+*
+TTRT_TRT TRT   0(*-*,6),0(8)       First match using translate table
+*
+          DS    0D
+***********************************************************************
+*         TRT tests; TTRTDS DSECT overlays a test
+***********************************************************************
+*
+TRTTests  DC    A(TRTT1,TTRTDSLN,TRTTn,0)  A(1st),len 1,A(last),unused
+*
```
abekornelis, line 477: There is no test using StrBeg - I'm sure you did that on purpose, but could you please add some comment to explain why this edge case is irrelevant?

rt/mlc/IS596.MLC
```diff
@@ -0,0 +1,4364 @@
...
+         LTORG ,
+*
+TTRTR_TRTR TRTR 0(*-*,9),0(8)      Reverse 1st match via translate tbl
+*
+          DS    0D
+***********************************************************************
+*         TRTR tests; TTRTDS DSECT overlays a test
+***********************************************************************
+*
+TRTRTests DC   A(TRTRT1,TTRTDSLN,TRTRTn,0) A(1st),len 1,A(last),unused
```
abekornelis, line 859: Test with StrEnd is missing, please explain why?

rt/mlc/IS596.MLC
```diff
@@ -0,0 +1,4364 @@
...
+***********************************************************************
+*         Put new tests above this line
+***********************************************************************
+TRTRTn    EQU  *-TTRTDSLN          Last test
+*
+          DS    0D
+***********************************************************************
+*         TRTR tests; 31-bit only; preset, expected GR1 bit 32 = 0
+***********************************************************************
+*
```
abekornelis, line 911: Tests for StrEnd and StrNone are missing, please explain why?

abekornelis (commented):
@jyganci No remarks - I like your table-driven approach to test case handling.
I did add a few questions, though. I'll be happy to approve, whatever your replies may be.

rt/mlc/IS596.MLC
jyganci: The pz390.java code is sloppy about going beyond of memory. It adds 32 bytes to the memory size during initialization so it can access beyond end of memory when checking fields. See the code there.

If I recall correctly, the statement at line 2598 was put in to avoid that. At the time the line was added, it was at the end of IS596.MLC. Now other data follows it, so the line can be deleted. If it's okay with you, I will just do this as part of the current PR. Okay?

rt/mlc/IS596.MLC
jyganci: StrBeg is used in the TRTR tests. The TRT tests only use StrNone, StrMid, and StrEnd. The TRTR tests only use StrNone, StrMid, and StrBeg. Don't see a need for having a test for the "first" byte to be a space (first = leftmost byte for TRT, rightmost byte for TRTR). If you feel strongly about this, I will add these tests. Let me know.

rt/mlc/IS596.MLC
jyganci: See comment I added for your TRT comment.

rt/mlc/IS596.MLC
jyganci: If you look at the #596 changes for the TRTR instruction, you will see that the setting of GR1 is done only if there is a space in the source string. The "31-bit only" tests were created specifically for the case a space is anywhere in the string, so using StrNone would not modify GR1. That was tested in the previous TRTR test set (GR1 bit 32 = 1). Don't see a need to do it again. As for not using StrEnd, see above TRT comments.

rt/mlc/IS596.MLC
abekornelis: that's fine with me.

rt/mlc/IS596.MLC
abekornelis: First position is (to my  mind) an edge case. One which I've seen fail (totally different context) in my own work. I would really appreciate if you're okay to add this test case.

rt/mlc/IS596.MLC
abekornelis: Same line of reasoning, same answer.

rt/mlc/IS596.MLC
abekornelis: For StrNone - agreed, no problem.
For StrEnd, again same reasoning, same answer (and same appreciation)

rt/mlc/IS596.MLC
jyganci: Added TRTR tests for StrEnd. Push coming shortly.

rt/mlc/IS596.MLC
jyganci: Deleted line. Push to clone coming shortly. Will also contain the TRT StrBeg tests and TRTR StrEnd tests for GR1 bit 32 equal to 1.  None for the TRTR GR1 bit 32 equal to 0.

rt/mlc/IS596.MLC
jyganci: TRT tests for StrBeg added. Push coming shortly.

abekornelis (approved):
Re-tested with latest changes. All okay. Thanks!

PR #871, opened 2026-07-09
Make rt/bash scripts executable

jyganci:
A few rt/bash scripts were accidentally not made executable
runmfacc
runscripts
runvsam1
runvsam2

Also, the RPI2006 test in runscripts  has been modified to have 'mem(32)' as a parameter.


abekornelis (dismissed):
You duplicated the RPI2006 fix so you'll have a conflict on that line of code.
But that change will undo the approval.

abekornelis (approved):
good to go ...

PR #872, opened 2026-07-10
Make MfAcc tests optional

### LINKED ISSUE #866: MFACC tests should be optional part of RT suite
The MFACC program collection has been made part of the regression test suite. But it is always being run, we'd rather have it be an optional component of each test run. Please find out how to make this subset of tests optional and make modifications to implement the change.

abekornelis:
Fixes #866


bash/help
```diff
@@ -18,6 +18,7 @@ echo "exec   - Execute a previously linked program"
 echo "help   - This help overview"
 echo "link   - Link a previously assembled or compiled program"
 echo "mac    - Run the macro processor"
+echo "runtest- Run the groovy test suite or a substet"
```
jyganci: "... or a substet". ??? Should this be subtest???

z390test/README.md
```diff
@@ -62,6 +59,24 @@ which provides methods for running and capturing the output.
...
+@Tag("optional")
+
+```
+
+See `RunMfAcc.groovy` for an example.
+
+Optional test script must be encoded explicitly in the `bat\RUNTEST.BAT`
+and the `bash/runtest` scripts because they require a slightly different invocation.
+
+Optional test cases are skipped when running a build procudere,
```
jyganci: "procudere" -> "procedure"

z390test/README.md
```diff
@@ -74,21 +89,21 @@ int rc = this.asm(basePath('tests', 'TESTINS1'), *options)
...
-* cblc - COBOL compile only
-* cblclg - COBOL compile, link and go. Any non-zero return will exist with RC
-* lz390 - Link only. Output not cleared prior to run
-* ez390 - Run only. Output not cleared prior to run
+- asm - Assemble only
+- asml - Assemble and link. Any non zero RC on assemble will exit with RC
+- asmlg - Assemble, link and go. Any non zero RC on assemble and link will exit with RC
+- mz390 - Assemble only
+- cblc - COBOL compile only
+- cblclg - COBOL compile, link and go. Any non-zero return will exist with RC
```
jyganci: There is also cblcl. Put it in the list???

jyganci:
I don't know enough gradle, groovy to verify the corresponding changes.

If I read this PR correctly, at the time of the PR there is only one optional test, mfacc. The runtest and RUNTEST.BAT scripts have that test name hardcoded. This implies to me that every time a new optional test is added, these two scripts will have to be unpdated. Correct?

While my quick look at the PR only found typos, someone with more gradle/groovy expertise should review the PR.

Finally, after looking at the PR, I now think this is a solution looking for a problem. The mfacc tests run quickly, so making them optional doesn't save but a few seconds. Just my opinion. Others may differ.

z390test/README.md
abekornelis: fixed.

z390test/README.md
abekornelis: done.

abekornelis:
@jyganci I'm looking at run time on the github server. Saves ca. 10% of the testjob's run time. This establishes a base - we may want to add more optional tests in the future. And as for hardcoding groovy script names - you are right that that is an ugly solution. I modified the bat/bash scripts to look for the actual marker in the target groovy script file

doc/contributing/contribute_code.md
```diff
@@ -106,6 +106,7 @@ The test scripts are in subdirectory z390test\src\test\groovy\org\z390\test
 | force a test run                          | `gradlew test --rerun`           |
 | run a specific test/testset               | `gradlew test --tests 'pattern'` |
 | run all zCobol tests                      | `gradlew test --tests '*cbl*'`   |
+| run optional tests                        | `gradlew optionalTest --rerun`   |
```
jyganci, line 109: Add line for run a specific optional test???

bash/runtest
```diff
@@ -0,0 +1,33 @@
...
+# For optional tests, a different test procedure must be specified
+testproc=test
+if [ -n "$1" ]; then
+  groovy_file="src/test/groovy/org/z390/test/${1}.groovy"
+  if [ -f "$groovy_file" ] && grep -qF '@Tag("optional")' "$groovy_file"; then
+    testproc=optionalTest
+  fi
+fi
+
+# execute the tests
```
jyganci, line 26: Should be "execute the test" ???

bash/runtest
```diff
@@ -0,0 +1,33 @@
...
+if [ $# -eq 0 ]; then
+  ./gradlew "$testproc" --rerun
+else
+  ./gradlew "$testproc" --rerun --tests "$@"
+fi
+
+exit $?
```
jyganci, line 33: Missing newline as last character ???

jyganci (changes requested):
I've left a few comments. Some might involve changes. I will approve after I see your response(s).

PR #882, opened 2026-07-25
Fix test programs for RPI 2009 and 2010

abekornelis:
there is no issue to describe this change

What I did:
- fix small issues with RT scripts for RPI2009 and RPI2010
- add fixed scripts to RunScripts (bat, bash, groovy)


rt/rt/RPI2009A.MLC
```diff
@@ -136,7 +136,7 @@ RPI20020 DS    0H
          ST    R1,WAPARM@              Save pointer to parameter list
          STMH  R12,R12,WAR12H          Save high half of R12
          LLGTR R12,R12                 Ensure high half of R12 is zero
-         MVC   WAEYE,=CL8'RPI2009A'    Set eyecatcher
+*        MVC   WAEYE,=CL8'RPI2009A'    Set eyecatcher
```
Zaneham, line 139: Was commenting the eyecatcher out deliberate, or was it to keep the addresses stable for the RT scripts? It's handy to have when reading a dump, so if it's only in the way because of the address matching it'd be nice to keep it and absorb the shift elsewhere.

rt/rt/RPI2009A_A.RT
```diff
@@ -3,7 +3,7 @@
 a 80d6.
 g
 s
-* ############################## (@228)
+* ############################## (mlc@200 @48 t85
```
Zaneham: Missing the closing paren. It's the only one out of the 63 `(mlc@...` comments added here that's short of it, so it looks like a slip rather than intent.

rt/bat/RUNSCRIPTS.BAT
```diff
@@ -18,68 +18,95 @@ call bat\asmlg %z_TraceMode% demo\hello      sysmac(mac) noloadhigh test(rtscrip
...
+call bat\asmlg %z_TraceMode% rt\rt\RPI1641   sysmac(mac) noloadhigh zvsam(2)                  || goto error
+
+call bat\asml  %z_TraceMode% rt\rt\RPI2009A  sysmac(mac) || goto error
+call bat\asml  %z_TraceMode% rt\rt\RPI2009B  sysmac(mac) || goto error
+call bat\asml  %z_TraceMode% rt\rt\RPI2009C  sysmac(mac) || goto error
+call bat\asml  %z_TraceMode% rt\rt\RPI2009D  sysmac(mac) || goto error
+call bat\asml  %z_TraceMode% rt\rt\RPI2009E  sysmac(mac) || goto error
+call bat\asml  %z_TraceMode% rt\rt\RPI2009F  sysmac(mac) || goto error
+
+set  rtscript=rt\rt\RPI2000.RT        && call bat\asmlg %z_TraceMode% rt\rt\RPI2000   sysmac(mac) noloadhigh             test(rtscript) || goto error
```
Zaneham, line 56: FYI rather than a request. `set rtscript=...` takes everything up to the `&&`, so the alignment spaces end up inside the value and `%rtscript%` is the path plus padding.

It's harmless here because z390 trims it (`get_ascii_env_var_string` in `src/sz390.java`, `text.trim()`, RPI111). Just worth knowing the column alignment quietly depends on that, in case the pattern gets reused for a variable that something else consumes.

Zaneham (commented):
Read through the whole thing. The RPI2009 additions look consistent and the bat/bash/groovy runners line up with each other. A few notes inline, none of them blocking.

rt/rt/RPI2009A.MLC
abekornelis: I agree that the eyecatcher is quite useful. However, shifting all locations by 6 bytes - except where alignment absorbs or extends - is a major time investment. I think (although I cannot remember) that I added the eyecatchers when importing the code into the repo, long ago. Then never looked at it anymore :-(
I'm not going to fix the offsets - more so because there is a (rather vague, admittedly) plan to create a ASSERT macro that will create the RT script during assembly. That would allow us to use symbolic addresses, rather than the hardcoded stuff we are now stuck with. It would greatly enhance the usability of the RT script facility. And would allow us to rewrite existing RT scripts by wiring them in with the actual assembler code being tested.
So yes, I agree - but decline to make any such changes right now.

rt/rt/RPI2009A_A.RT
abekornelis: Thanks for catching that one. I fixed it.

rt/bat/RUNSCRIPTS.BAT
abekornelis: Okay - thanks for informing me. I think I've been bitten by that one before. I'll try to remember. 
For readability, I much prefer a columnar format of repetitive code. I'm glad, at least in this instance, I can get away with my columnar obsession. That's what you get from doing SQL for over a dozen years :-)

Zaneham (approved):

PR #895, opened 2026-08-08
changing zc390 to prepare for the java linter

abekornelis:

What I did:
I ran the linter (in development; PR #806 ) to validate zc390, zACB, zCatVAIX, zCatVCLR
then fixed all flagged statements. The intention is not to change any code, just whitespace and placement of curly braces.


src/zACB.java
```diff
@@ -19,17 +19,21 @@
 
 
 
-/***************************************************************************
- * 
+/* *************************************************** Maintenance
```
Zaneham: This opens with `/* *** Maintenance` on the same line, while vz390 and tz390 put the banner on its own line. Worth making them match if you care. Same in zCatVAIX and zCatVCLR.

src/zCatVAIX.java
```diff
@@ -19,15 +19,20 @@
 
 
 
+/* *************************************************** Maintenance
+ * 2016-12-28 initial coding
+ * 2026-03-23 AFK Fix/Add javadoc comments
+ * 2026-08-06 AFK Fix linter issues
+ * 2026-08-07 #807 Fix issues flagged by linter
```
Zaneham: Two entries for the same work here. Same in zCatVCLR.

Zaneham (commented):
Hey Abe, no issues with the code here. I compiled base and head and diffed the javap output across every class, comes out identical, and checkstyle is clean on all four files. Two cosmetic things inline.

Zaneham (dismissed):

src/zACB.java
abekornelis: justifiable. Fixed.

src/zCatVAIX.java
abekornelis: ouch - copy-past error. Fixed.

Zaneham (approved):

PR #896, opened 2026-08-08
Fix issues reported by linter in vz390

abekornelis:

What I did: I used the linter under development in PR#806 and fixed all issues it flagged in vz390.java

### DIFF
--- a/src/vz390.java
+++ b/src/vz390.java
@@ -1305,7 +1305,7 @@ public class vz390 {
 
         cur_vclr_flag = pz390.mem.getInt(cur_acb_vclra + vclr_flag);
         cur_vclr_lrec = pz390.mem.getInt(cur_acb_vclra + vclr_lrec);
-        if (cur_vcdt_ptha != 0 && (cur_vpth_flag & vpth_flag_aixp) != 0){  // RPI 865 is476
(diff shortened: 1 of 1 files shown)


src/vz390.java
```diff
@@ -849,3966 +856,3948 @@ public class vz390 {
...
+            set_feedback(pdf_def, rc_log, cmp_ves, rn_rcd_not_fnd);
+            return;
+        }
+        set_rpl_lxrba(cur_vx0_xrba);
+        // read ves record at cur vs0 index
+        if (!read_ves_rec()) { // RPI 688 RPI 672
+            return;
+        }
+        if ((cur_rpl_opt & rpl_opt_gen) != 0) {
+            System.arraycopy(pz390.mem_byte, cur_rpl_area + cur_vclr_koff,
```
Zaneham: The `// RPI 779` that sat on the `System` line before the arraycopy got folded away here. Worth putting back, since it's on real code rather than a brace line.

Zaneham (commented):
Bytecode is identical to base so this one is safe. One marker has gone, flagged inline. The two split RPI markers you rejoined further down are a nice fix.

src/vz390.java
abekornelis: Okay - I put that comment back in. Thanks for noticing and alerting me.

Zaneham (approved):

PR #897, opened 2026-08-08
Fix issues reported by linter in tz390

abekornelis:

What I did: I used the linter under development in PR#806 and fixed all issues it flagged in tz390.java


Zaneham (commented):
I checked this one bytecode-first. I compiled base and head and read the disassembly myself to confirm nothing had moved, and used Claude to help sweep the rest of the file while I've still got a subscription until the 14th. Identical across every class, and checkstyle is clean, so no behaviour change anywhere.

One general thing. About 18 `#nnn` markers have gone where they sat on a bare brace line that got merged away. You carried 155 of them across, so I assume these are misses rather than deliberate. Not a blocker either way. Rest is inline.

src/tz390.java
```diff
@@ -53,316 +53,321 @@
...
+ * 09/05/15 RPI 1529  Incorrect character x'92' in ascii character translation tables (backquote, should be quote = x'27')
+ * 10/03/15 RPI 1533  Invalid codepage option is not flagged as an error. Should cause abortion
+ * 03/01/16 RPI 2003  Add support for LAM, LAMY, STAM, and STAMY instructions
+ * 12/24/16 RPI 1598  Provide a means to select either original VSAM or the new one
+ * 05/21/18 RPI 1606  Circular reference of options file causes unspecified exception in tz390
+ * 2019-09-22 RPI 2201 dsh fix depreciated static decode by changing to Integer.decode
+ * 2019-10-01 RPI 2202 dsh add new instructions up to z15 including NCRK, NCGRK, MVCRL
+ * 2020-08-29 RPI 2202 dsh TESTINS6.MLC complete with 2121 opcode+operand tests for POP SA22-7832-12
+ *                         Version V1.7.00 with z390.jar for Oracle SE 8 and Open JDK 11.0.08
+ * 2020-10-18 RPI 2202 DSH add missing mnemonics BI,CLT,CLGT,LOCHI,LOCGHI,LOCHHI,LOCFHR,STOC,STOCG,STOCFH
```
Zaneham, line 303: Good catch on this line, it had a tab and a caret where the leading star should have been.

src/tz390.java
```diff
@@ -877,434 +882,436 @@ public  class  tz390 {
...
+
+
+
+    /**
+     * dfp_exp_bcd_to_cf5 returns CF5 5 bit
+     * combination field using index made up of
+     * high 2 bits of bias exponent
+     * plus 4 bit BCDnibble for first digit.
+     */
+    byte[] dfp_exp_bcd_to_cf5 = {0x00,0x01,0x02,0x03,0x04,0x05,0x06,0x07,0x08,0x09,0,0,0,0,0,0, //0d // RPI 407 indexed by high 2 bits of exp + fisrt digit
```
Zaneham: The `RPI 407` comment described the whole table, but it's now after the `//0d` row comment so it reads as if it describes row zero. Better off in the javadoc above.

src/tz390.java
```diff
@@ -877,434 +882,436 @@ public  class  tz390 {
...
+                             3,               //30 infinity  RPI 536
+                             4};              //31 NaN      RPI 536
+
+
+
+    /**
+     * dfp_cf5_to_bcd returns decimal digit 0-9
+     * indexed by 5 bit combination field value
+     */
+    long[] dfp_cf5_to_bcd = {0,1,2,3,4,5,6,7,  //00-07 //cf5 value
```
Zaneham: Same thing here, `//cf5 value` was the table header and now trails `//00-07`.

src/tz390.java
```diff
@@ -877,434 +882,436 @@ public  class  tz390 {
...
+     * dfp_dpd_to_bcd returns 3 digit decimal
+     * value 0-999 using 10 bit densely packed
+     * decimal index value.
+     * Notes:
+     * <ol>
+     *  <li>Redundent values in (...)</li>
+     *  <li>Java interprets leading 08 as octal number like 0x is hex so any leading 0's should be removed</li>
+     * </ol>
+     */
+    long[] dfp_dpd_to_bcd = {0,    1,  2,  3,  4,  5,  6,  7,  8,  9, 80, 81,800,801,880,881,    // 00
```
Zaneham, line 1112: Same again, and the first row's alignment has drifted against the rows under it.

abekornelis:
@Zaneham "About 18 #nnn markers have gone where they sat on a bare brace line that got merged away."
Honestly - a change marker on a bare brace. I simply did not care enough to carry them forward. If you think it's worthwhile I'll be happy to put them back on the line that now has the brace. The original line carrying the brace no longer exists.

src/tz390.java
abekornelis: @Zaneham you are correct. I have moved the comment into the javadoc.
The change marker is dropped - I  don't like change markers in the javadoc as it is propagated to external docs.

src/tz390.java
abekornelis: @Zaneham Okay - I modified the wording too - for better clarity.

src/tz390.java
abekornelis: It's ugly - I tried inserting leading zeros, but that makes them octal, resulting in test failures. To satisfy the linter I have to left-align the first column. I've added an explanatory note.

Zaneham (approved):

Zaneham (approved):

PR #904, opened 2026-08-10
Fix issues flagged by linter in ez390 and ArchLevel

abekornelis:

What I did: I used the linter under development in PR#806 and fixed all issues it flagged in ez390 and ArchLevel


Zaneham (approved):
Read this and this looks good to me.

PR #905, opened 2026-08-11
Fix issues flagged by linter in az390

abekornelis:

What I did: I used the linter under development in PR#806 and fixed all issues it flagged in az390


Zaneham (approved):
Looking all good here. Nothing to add.

PR #917, opened 2026-08-27
Add missing break statement in various programs

### LINKED ISSUE #916: missing break statements in pz390
In pz390.trace_ins() for cases 410, 420, 430 (around line 21360 in current code) the break statements are missing, causing confusing output during tracing (trace data are overwritten, last section wins)
Please add the missing break statements.

With the help of AI a few more comparable issues were located. I'll fix those as well.

abekornelis:
fixes #916

### DIFF
--- a/src/mz390.java
+++ b/src/mz390.java
@@ -12145,8 +12145,10 @@ public  class  mz390 {
                 return "" + seta_value2;
             case 3:
                 return "" + seta_value;
+            default:                // #916
+                tz390.abort_case(); // #916
+                return "";          // #916
             }
-            // intentional drop-through to reach error handling under default #916
         case 2: // val_setb_type
             switch (parm) {
             case 1:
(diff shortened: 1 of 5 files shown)


src/mz390.java
```diff
@@ -12143,6 +12146,7 @@ private String get_pc_trace_val(byte type,int parm) {
             case 3:
                 return "" + seta_value;
             }
+            // intentional drop-through to reach error handling under default #916
```
jyganci: I'm confused. Outer switch is on type. If its first case (for type 1) applies and its inner switch, on parm, does not meet any of the case values (1,2,3), doesn't that mean that there is an error? Shouldn't each of the inner switches have its own default case?  If the code drops into type case 2, might the invalid parm value, supposedly for type case 1, be valid for case type 2? I assume you've thought this through and dropping through is okay.

abekornelis: @jyganci I noticed each major branch has the exact identical set of minor cases. Theoretically, if ever an unsupported minor comes if for major cases 1 or 2, where that minor case has been added as supported under major case 3 that might cause unwanted behavior. I have copied the default block. It's a tad ugly, but safer.

jyganci (approved):
The inner default switch cases don't look ugly to me. All looks good.

PR #918, opened 2026-08-28
Revive tests for command processor handling

abekornelis:
There is no issue describing these changes.

What I did:
- fixed a few issues with TESTCMD1 to make it run compatibly on Linux. (it was Windows-only)
- dropped TESTCMD2 - it was Windows-only and tested a subset of the TESTCMD4 tests
- TESTMD3 and TESTCMD4 were in the groovy script - I added them to the RUNCMDPROC.BAT and runcmdproc bash scripts
- Fixed CRLF line endings in called bash scripts


Zaneham:
One thing on the Linux side. `rt/bash/runcmdproc` is mode 644 while every other runner in `rt/bash` is 755, so running it directly gives permission denied. Looks like a `git add --chmod=+x` job, which is the row you just added to the contributing doc.

abekornelis:
@Zaneham thanks for spotting the omission - I have fixed it.

Zaneham (approved):

PR #925, opened 2026-09-03
update GUI to use BoxLayout and GridBagLayout

### LINKED ISSUE #923: Change z390 GUI layout manager
The z390 GUI uses Java's FlowLayout manager to create the z390 window. This results in some components moving if the user resizes the window. In particular, the "Command" and "Status" lines are each made up of two components -- a label ("Cpmmand" respectively "Status") and a text field. Note that the "lines" that are shown in the window are not really lines, but where Java arranges the components. When resizing the z390 window horizontally, these components are free to move so may end up on separate lines (be rearranged). In particular, the "Status" label may move to the "Command" line because additional horizontal space in the window makes room for the "Status" component to flow to the previous line. To prevent this, another Java layout manager should be used to layout the components. The GridBayLayout manager is probably the best bet.

jyganci:
Fixes #923. The layout manager for the z390.java main panel is changed from FlowLayout to BoxLayout and GridBagLayout. The View-Status menu check box item is no longer needed but is left in place as "checked" but is unable to be unchecked. The corresponding command, STATUS, is modified to ignore "ON" and "OFF" requests, and changes the way "STATUS dec" handles its parameter. If the parameter, dec, is a valid nonnegative integer, it is used to set the status value; zero sets the status to off, greater than sets the status to dec * 1000  -- note that this could result in an integer that exceeds the maximum positive integer, but that's what the existing code does, so it is left as is. If the dec value is not a valid nonnegative integer, an error message is shown; no change to the status value. Old behavior for this case turned the status off.


src/z390.java
```diff
@@ -1511,7 +1517,31 @@ private void set_text_font() {
 
 
     /**
-     * return int from immediate decimal parm
+     * Return non-negative int value from string of decimal digits
+     * @param cmd_parm string of decimal digits
```
abekornelis, line 1521: ..\..\src\z390.java:1521: error: @param name not found
     * @param cmd_parm string of decimal digits

src/z390.java
```diff
@@ -1511,7 +1517,31 @@ private void set_text_font() {
...
+     * @return integer value if string contains only decimal
+     *         digits and the parsed value is in the range
+     *         zero to Integer.MAX_VALUE; otherwise, returns -1
+     */
+    private int get_nonnegative_dec_int(String s) {            // #923
```
abekornelis, line 1526: ..\..\src\z390.java:1526: warning: no @param for s
    private int get_nonnegative_dec_int(String s) {            // #923

abekornelis (changes requested):
Please fix the javadoc at c. 1521.
Otherwise okay - passes all tests

PR #928, opened 2026-09-06
Revive 3 rt\test testcases

abekornelis:
This change is not related to any issue

What I did:
- scripted bat/bash/groovy tests for three more testcase programs in the rt\test folder
- added a reference file for one of them
- added a comparison step to each script to validate actual output against reference file


Zaneham (approved):
Looking good!

